PR #2, opened 2024-02-26
Add instructions to compile binaries

kavir1698:
Tests fail because the changes in #1 need to be merged first.


minottic (approved):

.github/workflows/go.yml
```diff
@@ -48,3 +48,71 @@ jobs:
...
+      - name: Get dependencies
+        run: go get -v ./...
+
+      - name: Build for Linux
+        run: |
+          PREFIX="build"
+          OS="linux"
+          APPS="datasetArchiver datasetGetProposal datasetRetriever datasetIngestor datasetCleaner"
+          
+          for GOOS in $OS; do
```
minottic, line 71: this loop is not looping

.github/workflows/go.yml
```diff
@@ -48,3 +48,71 @@ jobs:
...
+        run: |
+          $PREFIX="build"
+          $OS="windows"
+          $APPS="datasetArchiver", "datasetGetProposal", "datasetRetriever", "datasetIngestor", "datasetCleaner"
+          
+          foreach ($GOOS in $OS) {
+            $OUT="$PREFIX/$GOOS"
+            mkdir -p "$OUT"
+            foreach ($APP in $APPS){
+              $env:GOOS=$GOOS
```
minottic, line 108: why do these need to be env vars differently from the `build for linux` step

.github/workflows/go.yml
minottic, line 1: general (ignorant) comment: how does this differ from: https://github.com/paulscherrerinstitute/scicat-cli/blob/main/.github/workflows/go.yml#L28?

.github/workflows/go.yml
kavir1698: how do you know? what's wrong?

.github/workflows/go.yml
kavir1698: You mean the fact that here we're explicit on what we're building?

.github/workflows/go.yml
minottic: OS is single valued so, this is looping on a single value. What's the point?

.github/workflows/go.yml
minottic: yes. What's the advantage?

kavir1698: The extra control. It doesn't hurt. This is following the `build.sh`

kavir1698: [Artifacts are retained for 90 days by default.](https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period).
We just have to release the artifacts we want. I'll modify accordingly.

minottic: ah nice! What about [this](https://goreleaser.com/ci/actions/)? I think under the hoods it indeed creates the artifacts

kavir1698: But, artifacts are preserved only to a maximum of 90 days. So if we want longer availability of them, we might have to think about other solutions.

kavir1698:
@minottic the goreleaser is working. I have modified it to run in this branch just for tests. Before merging, we have to revert back to [47c59ed](https://github.com/paulscherrerinstitute/scicat-cli/pull/2/commits/47c59ed63ee14fea056cf85bfde0b6d73f8bc81c). After it is merged, we need to push a tag, and it should create a release with all the binaries.

kavir1698:
We have to tag the main branch after this PR is merged for the releases to be created. What would you choose as the version of the whole package? v1.0.0? or 0.1.0?

minottic:
I would do 0.1.0 for the moment

PR #4, opened 2024-02-26
Add test for checkMetadata function

kavir1698:
Once this test is merged, I will refactor the function and add checks to address #9


datasetIngestor/checkMetadata.go
```diff
@@ -18,7 +18,10 @@ import (
 const DUMMY_TIME = "2300-01-01T11:11:11.000Z"
 const DUMMY_OWNER = "x12345"
 
+// getHost is a function that attempts to retrieve and return the fully qualified domain name (FQDN) of the current host.
+// If it encounters any error during the process, it gracefully falls back to returning the simple hostname or "unknown".
 func getHost() string {
```
minottic, line 23: is this getHost ever used? if not, maybe we can remove in the next pr

datasetIngestor/export_test.go
minottic, line 1: what's this file for? can't we simply call getHost in the test?

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 
 require (
 	github.com/creack/pty v1.1.17 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
```
minottic, line 16: what's this?

datasetIngestor/checkMetadata_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+    // fail the test and report an error if the returned hostname is "unknown".
+    if host == "unknown" {
+        t.Errorf("getHost() was unable to get the hostname")
+    }
+}
+
+func TestCheckMetadata(t *testing.T) {
+	// Define mock parameters for the function
+	var TEST_API_SERVER string = "https://dacat-qa.psi.ch/api/v3" // "https://example.com/api"
```
minottic, line 27: I think we shouldn't connect to dacat for real, but rather fake the call response. If we do this, this test has no dependency on dacat and will work even if dacat will change address or if other RIs want to use the CLI

datasetIngestor/checkMetadata.go
kavir1698: It is used. See line 138

datasetIngestor/export_test.go
kavir1698: No. only functions whose name start with a capital letter are exported, and we need exported functions to call them in the test file.

go.mod
kavir1698: Automatically added as a result of imports

datasetIngestor/checkMetadata.go
minottic: my bad, sorry!

datasetIngestor/checkMetadata_test.go
kavir1698: The problem is the tests fail with not information if I change the server.

minottic: could it be that it fails [here](https://github.com/paulscherrerinstitute/scicat-cli/blob/3c7a584def4f406a306a2981611d18c8f081d127/datasetIngestor/checkMetadata.go#L228)? If so, and if there isn't a way to mock it in go, I think we could proceed with this PR and do a refactor after that, using in line 228 the client (which we can mock)

minottic: maybe in PR #17

kavir1698: > could it be that it fails [here](https://github.com/paulscherrerinstitute/scicat-cli/blob/3c7a584def4f406a306a2981611d18c8f081d127/datasetIngestor/checkMetadata.go#L228)?

I tested that too. I created local urls for all the cases `myurl` would be. It fails without any error message.

But yeah, we either can spend an unknown amount of time here, or just mark this part of the test to be improved in the future. Additionally, the APIServer is not an important part of the `CheckMetadata` function, it is used in `GetProposal`, for which we will have independent tests.

#17 just makes the code cleaner and more robust, the logic of the code remains the same.

Shall I mark this line as a TODO?

minottic: I think a TODO is fine, maybe you could open an issue for that, so we don't forget. I am fine as well to do it outside of #17, good point.

kavir1698: Ok, I'll create an issue after this is merged.

minottic (approved):

kavir1698:
There is one change request from @sbliven that's blocking the merge, but I can't see where that is 🤔

minottic:
Spencer has to approve the changes were made, otherwise it's blocked

sbliven (approved):
I'm not sure what the change I requested was, but this looks fine. Good tests.

PR #15, opened 2024-03-05
Test datasetIngestor/main.go prints help screen

kavir1698:
Was part of #1. After this and #14  are merged, #1 can be removed.


kavir1698:
> could you please try to see if [these function calls](https://github.com/paulscherrerinstitute/scicat-cli/blob/main/cmd/datasetIngestor/main.go#L143-L144) require an internet connection?
> 
> If so, could you please mock them? Otherwise we couple the tests of the CLI arguments with the resolution of the connection (I think these checks should have been called after the CLI arguments parsing when it was written, so they wouldn't be called if the instruction was to exist before)

Yes, the functions require internet access. They make an HTTP GET request to a specific URL. When there is no internet access, the HTTP GET request will fail. However, the functions handle this failure without crashing: it logs an error message and then returns. So no need to change them for this test

minottic (approved):

PR #17, opened 2024-03-25
Refactor checkMetadata

kavir1698:


datasetIngestor/checkMetadata.go
```diff
@@ -2,111 +2,203 @@ package datasetIngestor
...
-	// use type assertion to access f's underlying map
-	metaDataMap = metadataObj.(map[string]interface{})
-	beamlineAccount = false
 
+	// Use type assertion to convert the interface{} object to a map[string]interface{}.
+	metaDataMap := metadataObj.(map[string]interface{}) // `.(` is type assertion: a way to extract the underlying value of an interface and check whether it's of a specific type.
+	return metaDataMap, err
+}
+
+func checkIllegalKeys(metadata map[string]interface{}) bool {
```
sbliven, line 78: Should document that `true` means illegal. I think I would have expected the opposite.

sbliven (approved):
This looks good. Good tests. Refactoring looks fine.

datasetIngestor/checkMetadata.go
kavir1698: I changed the function name to `containsIllegalKeys` so that true makes sense for illegal.

Issue #10, opened 2024-02-27
Improve scicat-cli release process


sbliven:
Now that the CLI is public and independent from other tools, let's revisit the release process.
1. Single version number for all tools using semantic versioning.
2. Tag releases and attach binaries to the release
3.  Use this repo for version checks
4.  Change downstream processes to get binaries from here rather than PSI-specific locations
5. Add CD actions to help with this

PR #18, opened 2024-03-25
Update CheckForNewVersion to use the GitHub location

kavir1698:
Will address point 3 on #10


datasetUtils/checkForNewVersion.go
```diff
@@ -49,19 +62,19 @@ func CheckForNewVersion(client *http.Client, APP string, VERSION string, interac
...
-		log.Println("Use the following command to update:")
-		log.Println("curl -O " + DeployLocation + APP + ";chmod +x " + APP)
-		log.Fatalf("Execution stopped, please update program now.\n")
-	} else if version.Compare(latestVersion, VERSION, ">") {
+	
+	// Get the operating system name
+	osName := runtime.GOOS
+	
+	// Generate the download URL
+	downloadURL := fmt.Sprintf("%s/scicat-cli_%s_%s_x86_64.tar.gz", DeployLocation, latestVersion, strings.Title(osName))
```
minottic, line 70: does this work? when I copy the link from the release I see: 
https://github.com/paulscherrerinstitute/scicat-cli/releases/download/v0.1.0/scicat-cli_.0.1.0_Linux_x86_64.tar.gz

(notice the `tag` after /download/)

datasetUtils/checkForNewVersion.go
```diff
@@ -49,19 +62,19 @@ func CheckForNewVersion(client *http.Client, APP string, VERSION string, interac
...
+	
+	if majorLatest > majorCurrent || version.Compare(latestVersion, VERSION, ">") {
 		log.Println("You should upgrade to a newer version")
 		log.Println("Current Version: ", VERSION)
 		log.Println("Latest  Version: ", latestVersion)
 		log.Println("Use the following command to update:")
-		log.Println("curl -O " + DeployLocation + APP + ";chmod +x " + APP)
+		log.Printf("curl -L -O %s; tar xzf scicat-cli_.%s_%s_x86_64.tar.gz; chmod +x %s\n", downloadURL, latestVersion, strings.Title(osName), APP)
```
minottic, line 77: should we say to use WSL for windows?

minottic, line 77: I think compared to before we are now saving the new executable in a folder (the unzipped `scicat-cli_.%s_%s_x86_64.tar.gz`. We should either say that as a comment or move the executable where it was before

datasetUtils/checkForNewVersion_test.go
```diff
@@ -0,0 +1,102 @@
...
+	var buf bytes.Buffer
+	
+	func init() {
+		// Redirect the output of the logger to buf
+		log.SetOutput(&buf)
+	}
+	
+	func getLogOutput() string {
+		return buf.String()
+	}
```
minottic, line 102: missing new line

datasetUtils/checkForNewVersion_test.go
```diff
@@ -0,0 +1,102 @@
...
+	tests := []struct {
+		name           string
+		currentVersion string
+		mockResponse   string
+		expectedLog    string
+		}{
+			{
+				name:           "New version available",
+				currentVersion: "0.9.0",
+				mockResponse:   `{"tag_name": "1.0.0"}`,
```
minottic, line 50: should this be **v**1.0.0?

datasetUtils/checkForNewVersion_test.go
```diff
@@ -0,0 +1,102 @@
...
+				expectedLog:    "You should upgrade to a newer version",
+			},
+			{
+				name:           "No new version available",
+				currentVersion: "1.0.0",
+				mockResponse:   `{"tag_name": "1.0.0"}`,
```
minottic, line 56: same as above

datasetUtils/checkForNewVersion_test.go
```diff
@@ -0,0 +1,102 @@
...
+			
+			// Create a mock HTTP client
+			client := server.Client()
+			
+			// Call CheckForNewVersion
+			CheckForNewVersion(client, "test", tt.currentVersion, false)
+			
+			// Check the log output
+			logOutput := getLogOutput()
+			if !strings.Contains(logOutput, tt.expectedLog) {
```
minottic, line 83: very minor comment, why is this a `contain` rather than `equal`?

datasetUtils/checkForNewVersion_test.go
```diff
@@ -0,0 +1,102 @@
...
+}
+	
+	var buf bytes.Buffer
+	
+	func init() {
+		// Redirect the output of the logger to buf
+		log.SetOutput(&buf)
+	}
+	
+	func getLogOutput() string {
```
minottic, line 100: maybe for later, this seems to be useful for all tests, so maybe in test util file later?

datasetUtils/checkForNewVersion_test.go
```diff
@@ -0,0 +1,102 @@
...
+			CheckForNewVersion(client, "test", tt.currentVersion, false)
+			
+			// Check the log output
+			logOutput := getLogOutput()
+			if !strings.Contains(logOutput, tt.expectedLog) {
+				t.Errorf("Expected log message not found: %s", logOutput)
+			}
+			
+			// Clear the log buffer after each test
+			buf.Reset()
```
minottic, line 88: this could maybe be in a cleanup? https://pkg.go.dev/testing#B.Cleanup

datasetUtils/checkForNewVersion.go
```diff
@@ -49,19 +62,19 @@ func CheckForNewVersion(client *http.Client, APP string, VERSION string, interac
...
-		log.Fatalf("Execution stopped, please update program now.\n")
-	} else if version.Compare(latestVersion, VERSION, ">") {
+	
+	// Get the operating system name
+	osName := runtime.GOOS
+	
+	// Generate the download URL
+	downloadURL := fmt.Sprintf("%s/scicat-cli_%s_%s_x86_64.tar.gz", DeployLocation, latestVersion, strings.Title(osName))
+	
+	if majorLatest > majorCurrent || version.Compare(latestVersion, VERSION, ">") {
```
minottic, line 72: what does this version.compare do?

minottic (commented):
very nice! the tests are very nicely written IMHO!

datasetUtils/checkForNewVersion.go
```diff
@@ -2,44 +2,57 @@ package datasetUtils
 
 import (
 	"bufio"
-	"io"
+	// "io"
```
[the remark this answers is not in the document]
kavir1698: Sure

datasetUtils/checkForNewVersion.go
kavir1698: You're right, I'll fix it.

datasetUtils/checkForNewVersion.go
kavir1698: Windows doesn't need chmod. The rest of the command is now different for windows.

datasetUtils/checkForNewVersion_test.go
kavir1698: Doesn't really matter, the code strips the v.

datasetUtils/checkForNewVersion_test.go
kavir1698: Because the test is interested in whether `tt.expectedLog` appears anywhere in the `logOutput`, rather than checking if `logOutput` is exactly equal to `tt.expectedLog`.

datasetUtils/checkForNewVersion_test.go
kavir1698: Maybe, but this is also one liner. Does the job good enough.

datasetUtils/checkForNewVersion.go
kavir1698: Compares two version numbers in the format x.x.x, given the operator in the end.

datasetUtils/checkForNewVersion_test.go
minottic: yes, but the fact that it is stripped is part of the logic to test

datasetUtils/checkForNewVersion_test.go
minottic: ok. Even though it is a little weird that the setup is in the init while the teardown is in the function

datasetUtils/checkForNewVersion_test.go
minottic: ah ok so logOutput concatenates all the prints in one string?

datasetUtils/checkForNewVersion.go
minottic: ah ok, thanks. Doesn't this always cover this `majorLatest > majorCurrent` then?

kavir1698: Yeah right, I can remove the first term.

datasetUtils/checkForNewVersion_test.go
kavir1698: I could move 
```
    var buf bytes.Buffer
    log.SetOutput(&buf)
```
inside the `TestCheckForNewVersion` function. But if we were to add new tests in the future, we'd have to rewrite that.

minottic: logically the "afterTest" part should belong to the cleanUp, so I'd rather move that there, rather than moving the init inside the func. But if you prefer having it here, that is fine with me

kavir1698: I don't have any personal preference here, other than avoiding redundancy. I just don't see a way to avoid having the clean up inside the test function. If we have multiple tests, we want them to clean up before the next test.

minottic: that's what the cleanup should do as the init does, it's a builtin function of the go test module if I understand right, that gets automatically called after each test. But I am fine with this anyway (also because I am not sure it does this in go)!

kavir1698: Yeah but I think the `t.Cleanup()` method still needs to be called from within the test function. It can't be called from outside the test function, because it needs a reference to the `testing.T` instance for the current test.

datasetUtils/checkForNewVersion_test.go
kavir1698: Yes, that's correct. It is capturing all the log messages that were written during the execution of the test case.

minottic (approved):

Issue #22, opened 2024-03-27
Handle errors fetching latest version


sbliven:
# Problem Description

The latest github outage resulted in our version files being replaced by a webpage explaining the 404 error. When datacatIngestor tried to access the version check, this resulted in dumping the full HTML as an error.

# Solution

Any errors with the version check (either network errors or parsing errors) should result in warnings but not block the ingestion process.

sbliven:
Part of the solution may be switching to `slog` or another logging package. That way we can have proper level-based logging with some of the less user-friendly error messages hidden behind a `--verbose` flag.

PR #25, opened 2024-03-27
Test check for service availability

### LINKED ISSUE #23: Update `CheckForServiceAvailability` to read from Github
`CheckForServiceAvailability` checks the availability of the dataset ingestor service. It fetches a YAML file, parses it, and logs the service availability status. It should read the YAML file from Github now.

The function also needs documentation, refactoring, and unit tests.

kavir1698:
Will close #23

### DIFF
--- a/datasetUtils/checkForServiceAvailability.go
+++ b/datasetUtils/checkForServiceAvailability.go
@@ -28,7 +28,10 @@ type ServiceAvailability struct {
 	Qa         OverallAvailability
 }
 
+// CheckForServiceAvailability checks the availability of the dataset ingestor service.
+// It fetches a YAML file from a specified location, parses it, and logs the service availability status.
 func CheckForServiceAvailability(client *http.Client, testenvFlag bool, autoarchiveFlag bool) {
+	// Send a GET request to fetch the service availability YAML file
 	resp, err := client.Get(DeployLocation + "datasetIngestorServiceAvailability.yml")
 	if err != nil {
 		fmt.Println("No Information about Service Availability")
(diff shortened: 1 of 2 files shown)


datasetUtils/checkForServiceAvailability.go
```diff
@@ -28,48 +26,66 @@ type ServiceAvailability struct {
...
+
+// CheckForServiceAvailability checks the availability of the dataset ingestor service.
+// It fetches a YAML file from GitHubMainLocation, parses it, and logs the service availability status.
 func CheckForServiceAvailability(client *http.Client, testenvFlag bool, autoarchiveFlag bool) {
-	resp, err := client.Get(DeployLocation + "datasetIngestorServiceAvailability.yml")
+	s, err := getServiceAvailability(client)
 	if err != nil {
-		fmt.Println("No Information about Service Availability")
+		log.Printf("Failed to get service availability: %v", err)
 		return
```
sbliven, line 37: This passes all errors on to the user. Are these generally suitable error messages, or will this result in confusing errors. This is not critical code, so I don't think we need to print long scary errors (#22).

sbliven (commented):
The downtime notifications are a nice feature. Did this already exist before? I never knew to update the availability status.

I'm not sure that storing the file in the main branch is the best way to go. This feels like mixing code with devops, although it's temptingly straightforward.

Thanks for the tests!

kavir1698:
Yes this was already there, I just cleaned the code and changed paths to Github.

Where would you store the yml file?

datasetUtils/checkForServiceAvailability.go
kavir1698: It is better to have these error messages than not so that one can see where the problem was.
`log.Printf` prints the error message to `stderr`, whereas `fmt.Println` prints to `stdout`.

sbliven:
Maybe a separate branch?

Let's merge this as is and discuss next week.

sbliven (approved):

minottic:
I have the feeling as well this is added complexity for no real reason. A better way to do this, IMHO, would be to make the BE return the unaivalabiltity and log it here (or maybe just simply notify the users). TBH I would get rid of this logic, but indeed we can do in another PR.

Something else though. I think this requires the file `...available.txt` which is missing from the release. So this will always fail right? should we merge the one with it first?

kavir1698:
> think this requires the file `...available.txt` which is missing from the release.

Sorry I don't see what file you're talking about. It checks the service availability from `datasetIngestorServiceAvailability.yml`.

minottic:
yes, that one. Ah ok I see it now on github. I got confused by [this](https://github.com/paulscherrerinstitute/scicat-cli/pull/25/files#diff-7f282fba44a40553199377927f9a3ebd9c3feab8ee651cf3ecd6ee71b4fe4fdeR23) and was looking for it in the release

kavir1698:
Ok then I'll merge this one and later we can talk about hosting the yml file on a different branch.

PR #29, opened 2024-04-03
Use the `nonInteractiveFlag` flag in `CheckForNewVersion` function calls

### LINKED ISSUE #8: datasetArchiver stops despite noninteractive flag
If the datasetArchiver is not on the most recent version, it stops despite the noninteractive flag.

kavir1698:
close #8


minottic (commented):
nice! I would add some tests to prove it fixes the issue

minottic (approved):
thx!

PR #34, opened 2024-04-04
Add golangci-lint to workflows

### LINKED ISSUE #30: Add Linting in CI
See [golangci-lint](https://github.com/golangci/golangci-lint)

kavir1698:
Closes #30

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,3 +65,17 @@ linters-settings:
     # Treat code as a program (not a library) and report unused exported identifiers; default is false.
     # The argument specifies the Go source directory to analyze.
     check-exported: false
+
+output:
+  formats:
+    - format: colored-line-number
+      path: stdout
+  print-issued-lines: true
+  print-linter-name: true
+  uniq-by-line: true
+  sort-results: true
+  sort-order:
+    - file
+    - linter
+    - severity
+  show-stats: true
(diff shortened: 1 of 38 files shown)


.golangci.yml
```diff
@@ -0,0 +1,71 @@
...
+
+  # Exit code when at least one issue was found, default is 1
+  issues-exit-code: 1
+
+  # Include test files or not, default is true
+  tests: true
+
+linters:
+  # Enable specific linters
+  enable:
```
minottic, line 13: as discussed, just to note it down, going forward we might want to relax these lists and enable further checks. For now I would say it's fine. Thanks!

.golangci.yml
```diff
@@ -0,0 +1,71 @@
+run:
+  # Timeout for analysis, e.g. 30s, 5m, default is 1m
+  deadline: 3m
```
minottic, line 3: do we need to increase the default?

minottic (commented):
thansk! 2 very minor comments...

.golangci.yml
kavir1698: Right. But some of the linters suggested things that didn't make sense. I tested each and every one and kept the ones whose errors could be fixed or whose error was important.

.golangci.yml
kavir1698: 1m sounds too short. I already see errors after 28 seconds.

minottic: yes, makes sense. Just something to keep in mind and maybe review later

minottic (approved):

PR #40, opened 2024-04-17
Remove the condition that ignores files in `AssembleFilelisting`

### LINKED ISSUE #39: File enumeration issue in v2.1.8
Tobias reports an issue with the 2.1.8 datasetIngestor deployed [here](https://gitlab.psi.ch/scicat/tools/-/tree/1f4148a390393c67623dfa8418cda5a025866d68). 

> It generated a dataset but only the folders are listed as data, not the actual files.

The scicat/tools version has been [rolled back](https://gitlab.psi.ch/scicat/tools/-/tree/7e00a520b9c7bc27a3aa8401a5ffedda8325336b) but presumably still exists in the github v2.1.8 release.

kavir1698:
Fixes #39

### DIFF
--- a/datasetIngestor/assembleFilelisting.go
+++ b/datasetIngestor/assembleFilelisting.go
@@ -46,6 +46,28 @@ func readLines(path string) ([]string, error) {
 	return lines, scanner.Err()
 }
 
+/*
+AssembleFilelisting scans a source folder and optionally a file listing, and returns a list of data files, the earliest and latest modification times, the owner, the number of files, and the total size of the files.
+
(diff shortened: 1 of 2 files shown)


minottic (commented):
since this is such a fundamental check, I would add some tests.

kavir1698:
> since this is such a fundamental check, I would add some tests.

Sure, but we need to decide how to check on the discovery #41

minottic:
can the first test be that it passes that point?

minottic (approved):

PR #43, opened 2024-04-22
Check rsync version

### LINKED ISSUE #42: Check rsync version
Ingestor is not working on RHEL9.

RHEL9 use rsync version 3.2.3 while RHEL8 use 3.1.3. They changed some commandline options between this two versions.
Ingestor currently execute on RHEL8 >`rsync -e ssh -q -avxz --progress --msgs2stderr  /path....`

This is not working anymore on RHEL9

The command line which work on RHEL9 > `rsync -e ssh -avxz --progress --stderr=error  /path....`

So the flag `-q` after ssh need to be removed and the `--msgs2stderr` changes to `--stderr=error`

kavir1698:
Will fix #42

### DIFF
--- a/datasetIngestor/syncDataToFileserver_unix.go
+++ b/datasetIngestor/syncDataToFileserver_unix.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	version "github.com/mcuadros/go-version"
 )
 
 // functionality needed for "de-central" data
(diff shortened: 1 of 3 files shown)


datasetIngestor/syncDataToFileserver_unix.go
```diff
@@ -9,6 +9,8 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	version "github.com/mcuadros/go-version"
```
minottic, line 12: is this needed by this PR?

datasetIngestor/syncDataToFileserver_unix.go
```diff
@@ -21,27 +23,49 @@ func SyncDataToFileserver(datasetId string, user map[string]string, RSYNCServer
...
 	}
-	// show rsync's output
-	// cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-
-	log.Printf("Running %v.\n", cmd.Args)
-	err = cmd.Run()
+	
+	// Check rsync version and adjust command accordingly
+	var rsyncCmd *exec.Cmd
```
minottic, line 33: this feels a little as code duplicaction. To me what changes is only that in some cases we have to do `ssh -q`. What if we maybe just make this a variable and simplify the IF?

minottic: (same for the --files-from, though slightly more complex, so we can keep it as is

minottic (commented):
going forward we could see if there is a go rsync wrapper. 

Would it be easy to add tests on the functions that use rsync to see if they still fail?

datasetIngestor/syncDataToFileserver_unix.go
kavir1698: > this feels a little as code duplicaction. To me what changes is only that in some cases we have to do ssh -q. What if we maybe just make this a variable and simplify the IF?

it both adding/removing `-q` and replacing `--stderr=error` and `--msgs2stderr`

kavir1698:
> going forward we could see if there is a go rsync wrapper.
> 
> Would it be easy to add tests on the functions that use rsync to see if they still fail?

After a little searching, I found a few rsync wrapper but they don't seem to be well maintained.

I tested the ingest manually and it worked. Testing `SyncDataToFileserver` would be a bit tricky with a lot of mocking.

datasetIngestor/syncDataToFileserver_unix.go
minottic: but at least for the version check, one could simply do: 

if version.Compare(versionNumber, "3.2.3", ">=") { ssh="ssh -q"} else {ssh="ssh}

and then just pass this variable outside of the if to the exec command

minottic: but at least for the version check, one could simply do:

if version.Compare(versionNumber, "3.2.3", ">=") { ssh="ssh -q"} else {ssh="ssh}

and then just pass this variable outside of the if to the exec command

kavir1698: I've change them now

minottic (approved):

datasetIngestor/syncDataToFileserver_unix.go
```diff
@@ -21,27 +23,52 @@ func SyncDataToFileserver(datasetId string, user map[string]string, RSYNCServer
...
 	if absFileListing != "" {
-		cmd = exec.Command("/usr/bin/rsync", "-e", "ssh -q", "-avxzr", "--progress", "--msgs2stderr", "--files-from", absFileListing, fullSourceFolderPath, serverConnectString)
+		rsyncFlags = append(rsyncFlags, "-r", "--files-from", absFileListing)
 	}
-	// show rsync's output
-	// cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	if version.Compare(versionNumber, "3.2.3", ">=") {
+		rsyncFlags = append(rsyncFlags, "--stderr=error")
+		rsyncCmd = exec.Command("/usr/bin/rsync", append(rsyncFlags, fullSourceFolderPath, serverConnectString)...)
```
minottic, line 40: sorry, I missed this yesterday... since the rsyncCmd is the same in the `if` and `else` I think we can move it after the block.

If we do this, it will become easy (I guess) to do the rsyncFlags build in a function which can later be tested

kavir1698: You didn't miss it yesterday, I just changed the code as I noticed a mistake in the code.

kavir1698: It is now a separate function and fully tested

minottic: thx!

minottic (approved):

datasetIngestor/syncDataToFileserver_unix.go
```diff
@@ -21,27 +23,54 @@ func SyncDataToFileserver(datasetId string, user map[string]string, RSYNCServer
...
+func buildRsyncCmd(versionNumber, absFileListing, fullSourceFolderPath, serverConnectString string) *exec.Cmd {
+	rsyncFlags := []string{"-e", "ssh", "-avxz", "--progress"}
+	if absFileListing != "" {
+		rsyncFlags = append([]string{"-r", "--files-from", absFileListing}, rsyncFlags...)
+	}
+	if version.Compare(versionNumber, "3.2.3", ">=") {
+		rsyncFlags = append(rsyncFlags, "--stderr=error")
+		// Full command: /usr/bin/rsync -e ssh -avxz --progress -r --files-from <absFileListing> --stderr=error <fullSourceFolderPath> <serverConnectString>
+	} else {
+		rsyncFlags = append(rsyncFlags, "-q", "--msgs2stderr")
```
sbliven, line 71: My mac has v 3.1.3 but doesn't support `--msgs2stderr`.

sbliven (approved):
Worked for me in -testenv from pmod7 (springdale 7.9) and ra-l-005 (rhel 8).

datasetIngestor/syncDataToFileserver_unix.go
kavir1698: Maybe we should reduce the threshold from 3.2.3?

kavir1698:
Since it wasn't working on RHEL9 before, and now the problem is just not printing the output, and since Spencer needs to have the version here up to 2.2.0, I think it is safe that we merge this now and fix the printing problem in a later PR

datasetIngestor/syncDataToFileserver_unix.go
kavir1698: I'm not sure what to do here.

sbliven:
Seems reasonable. 🚀

datasetIngestor/syncDataToFileserver_unix.go
sbliven: Let's leave it as an open issue

PR #45, opened 2024-04-22
Replace `io/ioutils` as it is deprecated

kavir1698:


kavir1698:
These are not covered by tests. But they are safe changes as the deprecated function is simply calling the new function.

PR #46, opened 2024-04-22
Refactor and test `AddAttachment`

kavir1698:


datasetIngestor/addAttachment.go
```diff
@@ -11,27 +11,28 @@ import (
...
+func ReadAndEncodeImage(attachmentFile string) (string, error) {
 	imgFile, err := os.Open(attachmentFile)
 	if err != nil {
-		log.Fatalf("Can not open attachment file %v \n", attachmentFile)
+		return "", err
 	}
-
 	defer imgFile.Close()
-
+
```
minottic, line 20: I wonder why the linting doesn't complain here

minottic (approved):
very minor comment. Looks good!

datasetIngestor/addAttachment.go
minottic: I thought this wasn't just an empty line, but an empty line with an extra `tab`

kavir1698: it is with a tab. But all my code is like that. That's how vscode does it by default

minottic: ok, that's weird, I thought lint would complain about that. Well not a problem anyway...

PR #50, opened 2024-04-24
Test `sendIngestCommand`

kavir1698:


datasetIngestor/sendIngestCommand.go
```diff
@@ -34,112 +47,144 @@ func createOrigBlock(start int, end int, filesArray []Datafile, datasetId string
 		totalSize += filesArray[i].Size
 	}
 	// fmt.Printf("Start:%v, end:%v, totalsize:%v\n, first entry:%v\n", start, end, totalSize, filesArray[start])
-
+
```
minottic, line 50: extra tab

datasetIngestor/sendIngestCommand_test.go
```diff
@@ -0,0 +1,177 @@
...
+		},
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			// Customize how redirects are handled if needed
+			// For a dummy client, default behavior is usually sufficient
+			return http.ErrUseLastResponse // Use the last response for redirects
+		},
+	}
+
+	// Mock user map
+	user := map[string]string{
```
minottic, line 59: is the user (and the later metadata) needed since you mock the HTTP client?

datasetIngestor/sendIngestCommand_test.go
minottic, line 1: the createOrigDatablocks tests are missing I think

minottic (commented):
there are quite some extra tabs. Maybe we can review the linting conf

datasetIngestor/sendIngestCommand.go
kavir1698: I don't think this is wrong because this is what happens when vscode reindents the code.

datasetIngestor/sendIngestCommand_test.go
kavir1698: Added

datasetIngestor/sendIngestCommand_test.go
kavir1698: user is a function argument.

minottic: sorry that was badly written. I wanted to say that user and metadata can have any value here as they are not really used later on, if I understand the test right

datasetIngestor/sendIngestCommand_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	user := map[string]string{
+		"accessToken": "testToken",
+	}
+	
+	// Call the function with test data
+	createOrigDatablocks(client, server.URL, datafiles, "testDatasetId", user)
+	
+	// Check if the correct number of requests were made
+	expectedRequestCount := (len(datafiles) + BLOCK_MAXFILES - 1) / BLOCK_MAXFILES
```
minottic, line 227: isn't `expectedRequestCount` equal to 1? It would be nice to have test cases for different values of expectedRequestCount (so changing BLOCK_MAXFILES) or changing the datafiles len. (if not too much work)

datasetIngestor/sendIngestCommand_test.go
kavir1698: Yes, their value is not important.

datasetIngestor/sendIngestCommand_test.go
kavir1698: It is added

datasetIngestor/sendIngestCommand_test.go
```diff
@@ -0,0 +1,258 @@
...
+	// Define test cases
+	testCases := []struct {
+		name           string
+		blockMaxFiles  int
+		datafiles      []Datafile
+		expectedRequests int
+		}{
+		{
+			name:          "Case 1: BLOCK_MAXFILES > len(datafiles)",
+			blockMaxFiles: 20000,
```
minottic, line 189: is `blockMaxFiles` ever used? (if so, I don't understand where, sorry)

datasetIngestor/sendIngestCommand_test.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+}
+
+func makeDatafiles(numFiles, size int) []Datafile {
+	datafiles := make([]Datafile, numFiles)
+	for i := range datafiles {
+		datafiles[i] = Datafile{Size: int64(size)}
+	}
+	return datafiles
+}
```
minottic, line 258: missing line

datasetIngestor/sendIngestCommand.go
minottic: mmmm vscode does the reindent because it has a default linter. But maybe we should review later on the linter settings and agree on a linter

datasetIngestor/sendIngestCommand_test.go
kavir1698: you're right. It is not used because it is a constant in the original code. Since it won't change anyways, I will just remove it from here.

minottic: but then how can test `#2` pass with `expectedRequests=2?`

kavir1698: It has to do with the `Size`.

minottic: ah right, thanks!

datasetIngestor/sendIngestCommand_test.go
minottic: should we simplify it then? (happy to keep it as is otherwise)

kavir1698: Sure

minottic (approved):
sorry I was so pedantic...

PR #51, opened 2024-04-24
Test `updateMetaData`

kavir1698:


datasetIngestor/updateMetaData_test.go
```diff
@@ -0,0 +1,162 @@
...
+		"ownerGroup": "testOwner",
+	}
+	tapecopies := new(int)
+	*tapecopies = 1
+	
+	// Call the function
+	updateClassificationField(client, APIServer, user, metaDataMap, tapecopies)
+	
+	// Check results
+	if _, ok := metaDataMap["classification"]; !ok {
```
minottic, line 159: can we please also check the value? and potentially if it changes with tape=2? (also, but up to you, would it make sense to have a dedicated function for this: https://github.com/paulscherrerinstitute/scicat-cli/pull/51/files#diff-84473b2b26cf0b84892987a98a715b48ecc9a1e070a19ab030595d1634a12d16R119-R157?)

kavir1698: I did that

minottic (approved):

PR #57, opened 2024-04-30
Test `CreateJob`

kavir1698:


datasetUtils/createJob_test.go
```diff
@@ -0,0 +1,99 @@
+package datasetUtils
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestCreateJob(t *testing.T) {
```
minottic, line 9: would it be easy to test if client.Do is called with the expected payload?

minottic (approved):
thanks!

PR #59, opened 2024-05-06
Test and refactor `CreateRetrieveJob`

kavir1698:


datasetUtils/createRetrieveJob_test.go
```diff
@@ -0,0 +1,145 @@
...
+		t.Errorf("constructJobRequest() returned invalid JSON: %v", err)
+	}
+
+	// Check if the JSON byte array contains the user email and username
+	if !strings.Contains(string(bmm), user["mail"]) {
+		t.Errorf("constructJobRequest() did not include the user email in the request")
+	}
+	if !strings.Contains(string(bmm), user["username"]) {
+		t.Errorf("constructJobRequest() did not include the username in the request")
+	}
```
minottic, line 69: could we also check that the JSON contains the datasetList? (would it be easier to check if the JSON, as a whole, equals what we expect?

kavir1698: Now checked

PR #79, opened 2024-06-05
Feature/cobra cli

### LINKED ISSUE #75: One executable for all CLI commands by using Cobra
# Proposal for a single executable based Scicat CLI tool
I would like to create a PR for a code rewrite that would put all of the separate executables defined in `cmd/` into one, where each component would get their own subcommand.

This would basically mean: Instead of calling the command this way:
```
./[command_name] (flags) (params)
```
We'd have the following convention:
```
./scicat-cli [command_name] (flags) (params)
```
This would be achieved through the [Cobra library](https://github.com/spf13/cobra/).

As an example of this concept, here's my [globus library](https://github.com/SwissOpenEM/globus) with a built-in cli implementation.

- [x] #79 
- [x] configure CI to allow for prereleases
- [ ] v3 -> main merging

consolethinks:
## Issue
Resolves #75 

## PR Description
It copies the pre-existing code in each executable's main function, adapts them to Cobra's way of getting flags and parameters. 

Each command supported by this one executable should behave the same way, for a given set of flags and parameters, as the corresponding tool that it was based on.

## Notes
This PR should be tested against PSI usecases still.


sbliven:
This introduces as `cmd/cmd` folder. This is a bit confusing. Should we rename the module?

consolethinks:
This last commit should address the double cmd folder structure by renaming the inner cmd folder to "commands" (as it contains command implementations of the CLI tool). If need be, the outer cmd folder could be renamed to cli as well, but I think the former is the "standard" in Go.

cmd/commands/datasetArchiver.go
```diff
@@ -1,143 +1,130 @@
...
-	tapecopies := flag.Int("tapecopies", 1, "Number of tapecopies to be used for archiving")
-	testenvFlag := flag.Bool("testenv", false, "Use test environment (qa) instead or production")
-	localenvFlag := flag.Bool("localenv", false, "Use local environment (local) instead or production")
-	devenvFlag := flag.Bool("devenv", false, "Use development environment instead or production")
-	nonInteractiveFlag := flag.Bool("noninteractive", false, "Defines if no questions will be asked, just do it - make sure you know what you are doing")
-	showVersion := flag.Bool("version", false, "Show version number and exit")
-
-	flag.Parse()
-
-	if datasetUtils.TestFlags != nil {
```
minottic: was this block removed by mistake?

cmd/main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import cmd "github.com/paulscherrerinstitute/scicat/cmd/commands"
+
+func main() {
```
minottic, line 5: could we test here that when we call this main with an arg (e.g. datasetIngestor) the datasetIngestor from `commands` is executed?

cmd/commands/datasetArchiver.go
minottic: having this will allow to check that passing variables by reference did not introduce any error

cmd/commands/datasetPublishData.go
```diff
@@ -1,349 +1,334 @@
...
-	Timeout:   10 * time.Second}
-var scanner = bufio.NewScanner(os.Stdin)
-var VERSION string
+		createWebpage := func(urls []string, title string, doi string, datasetDetails []datasetUtils.Dataset,
+			publishedDataId string, userpass *string, token *string) {
+			// A Function that returns the longest common prefix path (runes)
+			// from the array of strings
+			commonPrefix := func(arr []string) string {
+				// return shortest string, length given in bytes
+				findMinLength := func(arr []string) int {
```
minottic, line 103: why do these need to become internal functions? is this because the command is now a pointer? I am not a fun of internal functions, as they will be very hard to test later

cmd/commands/datasetRetriever.go
```diff
@@ -24,208 +10,205 @@ import (
...
+
+		var APIServer string = PROD_API_SERVER
+		var RSYNCServer string = PROD_RSYNC_RETRIEVE_SERVER
+
+		var client = &http.Client{
+			Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: false}},
+			Timeout:   10 * time.Second}
+
+		// internal functions
+		assembleRsyncCommands := func(username string, datasetDetails []datasetUtils.Dataset, destinationPath string) ([]string, []string) {
```
minottic, line 60: see comment above

minottic (commented):
I feel tests are missing and, once ready, I will be worried to merge all the main changes as a single commit. Also the CI breaks. 

This is a very useful PR, thanks!

cmd/commands/root.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "cmd",
+	Short: "CLI app for interacting with a SciCat instance",
+	Long: `This library comprises a few subcommands for managing SciCat
+and datasets on it, as well as interacting with the archival system connected
+to it.`,
+	// uncomment the next line if there's a default action
```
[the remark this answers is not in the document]
consolethinks: I might add a default action since the version parameter is common between all commands

cmd/commands/datasetArchiver.go
consolethinks: it was deleted because the code was original only copied from before I added tests, this is readded in the latest commit.

cmd/main.go
consolethinks: I have done something similar to that in the latest commit (calling Execute() instead of main, but it does effectively the same thing)

cmd/commands/datasetPublishData.go
consolethinks: they are internal because the functions are only used once and only exist to make the code more readable. If they weren't internal, the functions' definitions would be shared among all the commands.  This wasn't an issue when we had lots of little "main" packages, but now there's a bigger "cmd" package that ties everything together.

sbliven:
Not to hold up this merge even more, but are we ok with removing the old tools? I had imagined keeping the old executable around for backwards compatibility. After removing them we will have to maintain two branches until we drop support for the old cli names.

minottic (approved):

minottic:
I think we can double check the CI and the version comparison and use this eventually.
image
This will allow us to get rid of the v3 branch and merge everything into main, even before officially releasing. 

From memory, the release workflow in the CI overrides the "prerelease" and "latest" flags, so we should check that

PR #80, opened 2024-06-12
Feature/main tests

consolethinks:
adds parameter retrieval tests to the cli tools.

### DIFF
--- a/cmd/datasetArchiver/main.go
+++ b/cmd/datasetArchiver/main.go
@@ -14,9 +14,10 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"github.com/paulscherrerinstitute/scicat/datasetUtils"
 	"strings"
 	"time"
+
+	"github.com/paulscherrerinstitute/scicat/datasetUtils"
 
 	"github.com/fatih/color"
 )
@@ -51,7 +52,21 @@ func main() {
 	showVersion := flag.Bool("version", false, "Show version number and exit")
 
 	flag.Parse()
-	
+
+	if datasetUtils.TestFlags != nil {
+		datasetUtils.TestFlags(map[string]interface{}{
+			"user":           *userpass,
+			"token":          *token,
+			"tapecopies":     *tapecopies,
+			"testenv":        *testenvFlag,
+			"localenv":       *localenvFlag,
+			"devenv":         *devenvFlag,
+			"noninteractive": *nonInteractiveFlag,
+			"version":        *showVersion,
+		})
+		return
+	}
+
 	if *showVersion {
 		fmt.Printf("%s\n", VERSION)
 		return
--- a/cmd/datasetArchiver/main_test.go
+++ b/cmd/datasetArchiver/main_test.go
@@ -1,0 +1,82 @@
+package main
+
+import (
+	"flag"
+	"os"
+	"testing"
+
+	"github.com/paulscherrerinstitute/scicat/datasetUtils"
+)
+
+func TestMainFlags(t *testing.T) {
+	// test cases
+	tests := []struct {
+		name  string
+		flags map[string]interface{}
+		args  []string
+	}{
+		{
+			name: "Test without flags",
+			flags: map[string]interface{}{
+				"testenv":        false,
+				"devenv":         false,
+				"localenv":       false,
+				"noninteractive": false,
+				"version":        false,
+				"user":           "",
+				"token":          "",
+				"tapecopies":     1,
+			},
+			args: []string{"test"},
+		},
+		{
+			name: "Set all flags",
+			flags: map[string]interface{}{
+				"testenv":        true,
+				"devenv":         true,
+				"localenv":       true,
+				"noninteractive": true,
+				"version":        true,
+				"user":           "usertest:passtest",
+				"token":          "token",
+				"tapecopies":     6571579,
+			},
+			args: []string{
+				"test",
+				"--testenv",
+				"--devenv",
+				"--localenv",
+				"--noninteractive",
+				"--user",
+				"usertest:passtest",
+				"--token",
+				"token",
+				"--tapecopies",
+				"6571579",
+				"--version",
+			},
+		},
+	}
+
+	// running test cases
+	for _, test := range tests {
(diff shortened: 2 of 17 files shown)


cmd/datasetArchiver/main_test.go
```diff
@@ -0,0 +1,108 @@
...
+		if !passing {
+			t.Fail()
+		}
+	}
+
+	os.Args = []string{"test"}
+	main()
+}
+
+func TestMainFlagSetting(t *testing.T) {
```
minottic, line 53: minor comment, up to you if you want to change. One could use loops on tests [here](https://github.com/paulscherrerinstitute/scicat-cli/blob/main/datasetUtils/authenticate_test.go#L33-L79) rather than having multiple functions. If we do this, we could even test more combinations/flags

cmd/datasetArchiver/main.go
```diff
@@ -14,15 +14,24 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"github.com/paulscherrerinstitute/scicat/datasetUtils"
 	"strings"
 	"time"
 
+	"github.com/paulscherrerinstitute/scicat/datasetUtils"
+
```
minottic, line 21: extra line. Is this on purpose? then should there be one before as well (I am guessing some sort of PEP8 for go?)

cmd/datasetIngestor/main_test.go
```diff
@@ -11,6 +12,8 @@ import (
 // It captures the stdout, runs the main function, and checks if the output contains the expected strings.
 // This just checks if the main function prints the help message.
 func TestMainOutput(t *testing.T) {
+	flag.CommandLine = flag.NewFlagSet("main output", flag.ExitOnError)
```
minottic: not too sure I understand why this is needed

minottic (commented):
some minor comments. I have only commented the first commit (apart from 1 comment). The same comments would apply to all commits.

Thanks!

cmd/datasetArchiver/main_test.go
```diff
@@ -0,0 +1,108 @@
...
+	flag.CommandLine = flag.NewFlagSet("flag defaults", flag.ExitOnError)
+
+	testFlags = func(
+		userpass string,
+		token string,
+		testenvFlag bool,
+		devenvFlag bool,
+		nonInteractiveFlag bool,
+		showVersion bool) {
+		passing := true
```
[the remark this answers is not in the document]
consolethinks: I did it as such in order to be able to collect multiple errors in a single test, instead of failing on the first error.

cmd/datasetArchiver/main.go
consolethinks: it's the golang extension for vscodium/vscode. It autoformats everything everytime I hit Ctrl + S, so I think this is the official suggestion for formatting (and go formatting can be extremely specific, like not allowing for non-inline block starts).

cmd/datasetIngestor/main_test.go
consolethinks: Because the flag library is persistent, and if we call the main() function multiple times per test, and we're using the flags library for defining flags in the main function, then we'll end up redefining the same flags which will cause a panic during testing only. 

My second set of commits introduced this fix, which effectively hides/deletes the old flag definitions for each test.

cmd/datasetArchiver/main.go
```diff
@@ -14,15 +14,24 @@ import (
...
 	"github.com/fatih/color"
 )
 
 var VERSION string
 
+var testFlags func(
```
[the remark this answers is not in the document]
consolethinks: I'll look into this next week

cmd/datasetArchiver/main_test.go
consolethinks: sure that sounds easily feasible.

cmd/datasetArchiver/main.go
minottic: ah ok, thanks. Should we keep it for the last commit as well then?

cmd/datasetIngestor/main_test.go
minottic: ah ok, thanks. Should we keep it for the last commit as well then?

cmd/datasetArchiver/main_test.go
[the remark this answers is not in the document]
minottic: ofc... My bad

cmd/datasetIngestor/main_test.go
consolethinks: Nope, because for some reason, that caused another issue with this test, and the simplest solution I could think of is to just remove it. The test wasn't written by me btw. so I haven't looked much into it why it doesn't like to start with an empty list of flags.

cmd/datasetIngestor/main_test.go
```diff
@@ -70,232 +71,122 @@ func TestMainOutput(t *testing.T) {
...
+	// running test cases
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			flag.CommandLine = flag.NewFlagSet(test.name, flag.ExitOnError)
+			datasetUtils.TestFlags = func(flags map[string]interface{}) {
+				passing := true
 
-func TestMainFlagSetting(t *testing.T) {
-	flag.CommandLine = flag.NewFlagSet("flag setting", flag.ExitOnError)
+				for bFlag := range test.boolFlags {
```
minottic, line 165: why are the 3 types kept distinct? In the end, don't we only want to compare the values?

minottic (approved):
very minor comment, looks good thanks! 

if not too much work, could we organise this PR in commits so that there is 1 commit for main tested (+1 for the TestFlags)? After that, we can merge keeping these commits so to easily rollout later in case

cmd/datasetIngestor/main_test.go
consolethinks: I don't know if go lets programmers attempt to compare generic interface{} types. I could try doing that in order to combine the types, if you'd like!

minottic: don't bother! thanks!

PR #86, opened 2024-06-26
Feature: backwards compatibility

consolethinks:
# Backwards Compatibilty with V2
A set of changes that allow for backwards compatbility with v2 calling syntax. It is useful for cases where there are automation scripts that used to call the old executables.

As these automation solutions are usually done under Unix systems, these scripts were written in BASH, and are only provided for Linux and macOS.


README.md
minottic, line 1: could we maybe add where to download them from and what incompatible changes v3 introduced (apologies for the contradictory message before)?

minottic (commented):
Minor comment, good otherwise.

Thanks!

README.md
```diff
@@ -25,11 +25,23 @@ Tools are compiled for the following architectures:
 
 These can be cross-compiled from any system.
 
+### V3 Changes
+The separate executables (like `datasetIngestor`, `datasetRetriever`...) were combined into one `scicat-cli` executable, with each executable's features available as commands given as the first parameter to this executable. 
+
+These commands bare the same names as the former executables. For example, if you called `./datasetIngestor [flags]` before, now it's `./scicat-cli datasetIngestor [flags]`.
+
+ Furthermore, the use of single hyphen, multi-letter flags is now discontinued (as it does not follow convention).
```
minottic, line 33: sorry for yet another comment, I think this is not very clear, especially for non DEV users

can you maybe add an example? Thanks

minottic (approved):
thx!

PR #88, opened 2024-07-08
Improve code readability

consolethinks:
- added NOTE and TODO comments
- rewritten some code to make it more readable
- renamed a few variables and functions in datasetIngestor related code


cmd/commands/datasetCleaner.go
```diff
@@ -75,29 +75,24 @@ For further help see "` + MANUAL + `"`,
...
 		color.Set(color.FgRed)
 		log.Printf("You are about to remove a dataset from the === %s === data catalog environment...", env)
 		color.Unset()
 
-		pid := ""
-
-		if len(args) == 1 {
-			pid = args[0]
-		} else {
+		if len(args) < 1 || len(args) > 1 {
```
minottic, line 91: isn't this simply len(args) != 1?

cmd/commands/datasetGetProposal.go
```diff
@@ -61,30 +60,24 @@ For further help see "` + MANUAL + `"`,
...
 
-		ownerGroup := ""
-
 		//TODO cleanup text formatting:
-		if len(args) == 1 {
-			ownerGroup = args[0]
-		} else {
-			log.Println("invalid number of args")
-			return
+		if len(args) < 1 || len(args) > 1 {
```
minottic, line 77: len(args) != 1?

cmd/commands/datasetRetriever.go
```diff
@@ -168,12 +165,10 @@ For further help see "` + MANUAL + `"`,
 
 		destinationPath := ""
 
-		if len(args) == 1 {
-			destinationPath = args[0]
-		} else {
-			log.Println("invalid number of args")
-			return
+		if len(args) < 1 || len(args) > 1 {
```
minottic, line 168: len(args) != 1?

cmd/commands/datasetRetriever.go
```diff
@@ -184,29 +179,29 @@ For further help see "` + MANUAL + `"`,
...
-			fmt.Printf("\n\nNo datasets found on intermediate cache server.\n")
-			fmt.Println("Did you submit a retrieve job from the data catalog first ?")
-		} else {
-			// get sourceFolder and other dataset related info for all Datasets
-			datasetDetails, err := datasetUtils.GetDatasetDetails(client, APIServer, user["accessToken"], datasetList, ownerGroup)
-			if err != nil {
-				log.Fatal(err)
-			}
+			log.Printf("\n\nNo datasets found on intermediate cache server.\n")
+			log.Fatalln("Did you submit a retrieve job from the data catalog first ?")
```
minottic, line 183: this was just a print before, is it correct to be a fatal?

cmd/commands/datasetIngestor.go
```diff
@@ -147,8 +147,7 @@ For Windows you need instead to specify -user username:password on the command l
 		absFileListing := ""
 
 		if len(args) <= 0 || len(args) >= 3 {
-			log.Println("invalid number of args")
-			return
```
minottic: was this changed to fatal because now datasetIngestor is called by cobra and the return does not stop execution? Should all return be changed then (I guess though cobra doesn't do anything more than calling this, so the return wouldn't hurt?)

minottic (commented):
looks good, thanks, just few minor comments.

cmd/commands/datasetCleaner.go
consolethinks: it is actually, I was just on auto pilot when writing this code, I'll change it in a bit.

cmd/commands/datasetRetriever.go
consolethinks: same as above

cmd/commands/datasetIngestor.go
consolethinks: it was changed to fatal because it spares one line of code, and it's a fail condition anyways. I think it's the convention as well in Go, if we're not writing a library (in which case this would be a return WITH an error object and printing to stdout would be forbidden). In some of the other commands, this was already a Fatal( / ln / f) call.

cmd/commands/datasetGetProposal.go
consolethinks: same as above

cmd/commands/datasetRetriever.go
minottic: ah right, I did not see you removed the else too!

minottic (approved):

PR #91, opened 2024-07-22
Start migration guide for v2 to v3

sbliven:
Working document for v3 syntax


sbliven:
Should this file be added here or as part of the user guide?

PR #97, opened 2024-08-13
Refactor/sync data

consolethinks:
Refactors syncDataToFileserver:
 - redirect progress report to optional io.Writer parameter
 - remove logging

### DIFF
--- a/datasetIngestor/syncDataToFileserver_unix.go
+++ b/datasetIngestor/syncDataToFileserver_unix.go
@@ -7,7 +7,6 @@ package datasetIngestor
 import (
 	"fmt"
 	"io"
-	"log"
 	"os/exec"
 	"regexp"
 	"strings"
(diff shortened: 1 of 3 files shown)


datasetIngestor/syncDataToFileserver_unix.go
```diff
@@ -1,39 +1,42 @@
...
 	"strings"
+
 	version "github.com/mcuadros/go-version"
-	"regexp"
 )
 
 // functionality needed for "de-central" data
-func SyncDataToFileserver(datasetId string, user map[string]string, RSYNCServer string, sourceFolder string, absFileListing string) (err error) {
+// copies data from a local machine to a fileserver, uses RSync underneath
+func SyncLocalDataToFileserver(datasetId string, user map[string]string, RSYNCServer string, sourceFolder string, absFileListing string, cmdOutput io.Writer) (err error) {
```
minottic, line 19: what's the reason for this extra arg, cmdOutput?

minottic (commented):
minor comment + I would remove the commented logs

datasetIngestor/syncDataToFileserver_unix.go
minottic: is that used now?

consolethinks: It is used for outputting the rsync progress to the stdout, when using the datasetIngest command with the copy flag

minottic: but the file one is not used if I am not mistaken. Anyway, it doesn't harm

minottic (approved):

PR #99, opened 2024-08-13
Refactor/check metadata

consolethinks:
Refactors checkMetadata:
 - improvements
 - error handling instead of logging
 - render some internal functions public for better flexibility down the line


minottic (commented):
I think this additional functions, as they do non trivial stuff, are worth testing

consolethinks:
There are no new additional functions, and they're implicitly tested by the main test case. I can write tests for them, but at that point those will be new additions.

minottic:
you are right, my bad

minottic (approved):

sbliven (approved):

PR #100, opened 2024-08-13
Refactor/check central data

consolethinks:
Refactors checkDataCentrallyAvailable:
 - error handling instead of logs
 - ssh command's output is output into an io.Writer parameter
 - error message improvements


minottic (approved):
I would remove the commented logs, otherwise looks good, thanks!

PR #102, opened 2024-08-22
OpenEM: Add ingestor app

phwissmann:
This is a cleaned up version of the ingester app I prototyped a while ago. 

The three commits show the journey from [wails.io](wails.io) framework to the ingester framework with task queue. It should help to understand how to use wails. Each of the commits should build and compile.

What's missing from the top of my head:
- webserver and REST API
- configuration
- running headless mode without UI needed to run as service (needs some some abstraction on the `runtime.emitEvent` interface)
- ingestion task implementation (WIP, have a draft)
- s3 transfer
- globus transfer
- tests

Not meant to be merged at this stage.

image


openem-ingestor/core/taskqueue.go
```diff
@@ -0,0 +1,163 @@
...
+		f := value.(IngestionTask)
+		if f.Cancel != nil {
+			f.Cancel()
+		}
+		w.datasetSourceFolders.Delete(id)
+		runtime.EventsEmit(w.AppContext, "folder-removed", id)
+	}
+}
+
+func (w *TaskQueue) ScheduleTask(id uuid.UUID) {
```
consolethinks, line 121: "schedule" kind of implies that it is to be done in the future at a specified date. I feel like "queue" or "enqueue" would be the better wording here. Though with it already being part of a struct called "TaskQueue" that would be confusing as well... TaskQueue could also become "TaskRequests" since the requests are not in a queue yet. In general, I feel like the naming in general for TaskQueue and related functions could be improved

openem-ingestor/core/taskqueue.go
```diff
@@ -0,0 +1,163 @@
...
+func (w *TaskQueue) IngestDataset(task_context context.Context, task IngestionTask) TaskResult {
+	start := time.Now()
+	// TODO: add ingestion function
+	// dataset_id, err := IngestDataset(task_context, w.AppContext, task)
+	time.Sleep(time.Second * 5)
+	datasetPID := "1"
+	end := time.Now()
+	elapsed := end.Sub(start)
+	return TaskResult{Dataset_PID: datasetPID, Elapsed_seconds: int(elapsed.Seconds()), Error: nil}
+}
```
consolethinks, line 163: Otherwise, it's a nice prototype!

openem-ingestor/desktop-app/app.go
```diff
@@ -0,0 +1,67 @@
...
+	a.taskqueue.CancelTask(uuid.MustParse(id))
+}
+func (a *App) RemoveTask(id string) {
+	a.taskqueue.RemoveTask(uuid.MustParse(id))
+}
+
+func (a *App) ScheduleTask(id uuid.UUID) {
+
+	a.taskqueue.ScheduleTask(id)
+}
```
consolethinks, line 67: very self explanatory, once you understand the task system a little! I like this

openem-ingestor/desktop-app/frontend/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+If you have state that's important to retain within a component, consider creating an external store which would not be
+replaced by HMR.
+
+```ts
+// store.ts
+// An extremely simple external store
+import { writable } from 'svelte/store'
+export default writable(0)
+```
```
consolethinks, line 65: Stating the obvious again, but this should be changed later

openem-ingestor/desktop-app/frontend/src/App.svelte
```diff
@@ -0,0 +1,144 @@
...
+  .input-box .input:hover {
+    border: none;
+    background-color: rgba(255, 255, 255, 1);
+  }
+
+  .input-box .input:focus {
+    border: none;
+    background-color: rgba(255, 255, 255, 1);
+  }
+</style>
```
consolethinks, line 144: makes total sense

openem-ingestor/core/taskqueue.go
```diff
@@ -0,0 +1,163 @@
...
+	S3_Bucket   string
+	Md5checksum bool
+}
+
+type IngestionTask struct {
+	// DatasetFolderId   uuid.UUID
+	DatasetFolder     DatasetFolder
+	ScicatUrl         string
+	ScicatAccessToken string
+	TransferMethod    TransferMethods
```
consolethinks, line 40: maybe somewhere here a function pointer could be put in that starts/does the file transfer, whose function signature is the same for all transfer types through the use of the TransferOptions struct. The underlying function would then only use those parameters in TransferOptions which are relevant to the transfer at hand.

openem-ingestor/service/main.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func main() {
+	if err := core.ReadConfig(); err != nil {
+		panic(fmt.Errorf("Failed to read config file: %w", err))
+	}
+	log.Printf("Config file used: %s", viper.ConfigFileUsed())
+	log.Println(viper.AllSettings())
+
+}
```
consolethinks, line 18: good that this is a separate package, the REST API shouldn't depend from Wails. I also like that common functionality could be put into the "core" package

openem-ingestor/service/openem-ingestor-config.yaml
```diff
@@ -0,0 +1,13 @@
+Scicat:
+  Host: http://openem-scopem.ethz.ch:89/api/v3
+  AccessToken: "asdf"
+Transfer:
+  Method: S3
+  S3:
+    Endpoint: scopem-openem.ethz.ch:9000
```
consolethinks, line 7: If this and the host above are real endpoints, you should probably change them to something imaginary. This is only an example config after all!

consolethinks (commented):
I left notes on the code, nothing set in stone though, so feel free to give counterpoints

openem-ingestor/core/api.go
```diff
@@ -0,0 +1,32 @@
...
+
+type DatasetFolder struct {
+	Id         uuid.UUID
+	FolderPath string
+}
+
+// Select a folder using a native menu
+func SelectFolder(context context.Context) (DatasetFolder, error) {
+	dialogOptions := runtime.OpenDialogOptions{
+		DefaultDirectory: "",
```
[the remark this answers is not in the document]
phwissmann: I was thinking to add this in the config but felt a bit premature. I think for now this can be set for your local machine during development

openem-ingestor/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Live Development
+
+To run in live development mode, run `wails dev` in the project directory. This will run a Vite development
+server that will provide very fast hot reload of your frontend changes. If you want to develop in a browser
+and have access to your Go methods, there is also a dev server that runs on http://localhost:34115. Connect
+to this in your browser, and you can call your Go code from devtools.
+
+## Building
+
+To build a redistributable, production mode package, use `wails build`.
```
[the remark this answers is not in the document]
phwissmann: Let me update the readme in general

openem-ingestor/core/config.go
```diff
@@ -0,0 +1,61 @@
...
+		return config, err
+	}
+	return config, nil
+}
+
+func ReadConfig() error {
+	viper.SetConfigName("openem-ingestor-config") // name of config file (without extension)
+	viper.SetConfigType("yaml")
+
+	userConfigDir, _ := os.UserConfigDir()
```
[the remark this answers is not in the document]
phwissmann: true

openem-ingestor/core/taskqueue.go
phwissmann: This part of actually implementing the ingestion/transfer I would leave to you. I can give you the options that are needed for S3 but you seem to have a good idea on how to abstract that.

phwissmann:
- [ ] CI is missing

openem-ingestor/service/openem-ingestor-config.yaml
phwissmann: Sure. I used real config values in order to use it during development. The debug task copies this file next to the executable. I added my debug config file to userConfigDir()/openem-ingestor/ which takes precedent, so this can now just be an example.

openem-ingestor/core/config.go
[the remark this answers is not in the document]
phwissmann: Fixed. This userConfigDir/openem-ingestor takes precedence to the config next to the executable. Might be counterintuitive, what do you think?

openem-ingestor/desktop-app/frontend/src/App.svelte
phwissmann: Best of my frontend dev abilities XD

openem-ingestor/desktop-app/app.go
```diff
@@ -0,0 +1,67 @@
...
+		return
+	}
+
+	err = a.taskqueue.CreateTask(folder)
+	if err != nil {
+		return
+	}
+}
+
+func (a *App) CancelTask(id string) {
```
phwissmann, line 57: id is inconsistent with UUID below. Review.

openem-ingestor/desktop-app/app.go
phwissmann: I think at some point we need to think about how persistent these tasks should be; currently everything is stored in memory. The tasks will not survive any restarts. There are proper task queue implementations that allow storing task/results in databases (can also just be a file) but also for the service use case it might need to be a bit more reliable.

openem-ingestor/core/config.go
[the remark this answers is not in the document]
consolethinks: It might be a better idea to have it the other way around, because the user could get confused why their changes don't apply when they modify the config next to the executable. But honestly, it's not too big of a concern

openem-ingestor/core/taskqueue.go
```diff
@@ -0,0 +1,163 @@
...
+	Error           error
+}
+
+func (w *TaskQueue) Startup() {
+
+	w.inputChannel = make(chan IngestionTask)
+	w.resultChannel = make(chan TaskResult)
+
+	// start multiple go routines/workers that will listen on the input channel
+	for worker := 1; worker <= w.Config.Misc.ConcurrencyLimit; worker++ {
```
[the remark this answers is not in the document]
consolethinks: I admit I'm not too familiar with go routines yet, so my suggestion might not be the "go way"

openem-ingestor/core/config.go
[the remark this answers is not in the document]
phwissmann: changed priority.

openem-ingestor/core/config.go
```diff
@@ -0,0 +1,62 @@
...
+
+	viper.AddConfigPath("./")
```
consolethinks, line 57: forward slash only works on non-Windows systems and if the PWD is elsewhere it will not work anywhere. The PWD could be elsewhere if the user runs the executable from a different directory like so: `some/path/to/executable/ingestor-app`. It's better to use `os.Executable()` with `path.Dir()` to get an absolute path to the executable's directory

openem-ingestor/desktop-app/app.go
consolethinks: It would be better to save them to some persistent storage, but it's not a priority for now. If the transfer never completes due to the service crashing or restarting, they can re-request ingestion, maybe with an option to use a pre-existing entry in the scicat database (or use the sourceFolder to see if there's already one).

minottic:
just noticed this PR, any reason why this is supposed to become part of the scicat-cli repo and not another repo in openEm which uses the scicat-cli interface to connect?

PR #104, opened 2024-08-23
refactor/get local filelist

consolethinks:
This overhauls the local filelist gathering function:
 - The function now uses callbacks for symlink handling and filename checking
 - Better error handling
 - No unwanted logging from a library function
 - If no callbacks are provided, sensible defaults are used

### DIFF
--- a/cmd/commands/datasetIngestor.go
+++ b/cmd/commands/datasetIngestor.go
@@ -423,7 +423,7 @@ For Windows you need instead to specify -user username:password on the command l
 			log.Printf("Number of datasets not stored because of too many files:%v\nPlease note that this will cancel any subsequent archive steps from this job !\n", tooLargeDatasets)
 		}
 		color.Unset()
-		//datasetIngestor.PrintFileInfos() // TODO: move this into cmd portion
+
 		// print file statistics
 		if skippedLinks > 0 {
 			color.Set(color.FgYellow)
--- a/cmd/commands/datasetIngestor.go
+++ b/cmd/commands/datasetIngestor.go
@@ -237,7 +237,6 @@ For Windows you need instead to specify -user username:password on the command l
 		// a destination location is defined by the archive system
 		// for now let the user decide if he needs a copy
 
-		// now everything is prepared, prepare to loop over all folders
 		if nocopyFlag {
 			copyFlag = false
 		}
@@ -256,6 +255,11 @@ For Windows you need instead to specify -user username:password on the command l
 			}
 		}
 
+		var skippedLinks uint = 0
+		var illegalFileNames uint = 0
+		localSymlinkCallback := createLocalSymlinkCallbackForFileLister(&skip, &skippedLinks)
+
+		// now everything is prepared, prepare to loop over all folders
 		var archivableDatasetList []string
 		for _, datasetSourceFolder := range datasetPaths {
 			log.Printf("===== Ingesting: \"%s\" =====\n", datasetSourceFolder)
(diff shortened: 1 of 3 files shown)


cmd/commands/datasetIngestor.go
```diff
@@ -237,25 +237,30 @@ For Windows you need instead to specify -user username:password on the command l
...
 			case "keep":
-				skip = "kA"
+				skipSymlinks = "kA"
 			default:
-				skip = "dA" // default behaviour = keep internal for all
+				skipSymlinks = "dA" // default behaviour = keep internal for all
 			}
 		}
 
+		var skippedLinks uint = 0
```
minottic, line 258: the fact that file wide variables are changed in place feels a little prone to errors (since it's not idempotent). Given that you are refactoring GetLocalFileList, why not make it return these 2 extra values?

cmd/commands/datasetIngestor.go
```diff
@@ -410,7 +423,17 @@ For Windows you need instead to specify -user username:password on the command l
 			log.Printf("Number of datasets not stored because of too many files:%v\nPlease note that this will cancel any subsequent archive steps from this job !\n", tooLargeDatasets)
 		}
 		color.Unset()
-		datasetIngestor.PrintFileInfos() // TODO: move this into cmd portion
+		//datasetIngestor.PrintFileInfos() // TODO: move this into cmd portion
```
minottic, line 426: any reason to keep the comment?

minottic (commented):
comment + I would remove all commented code

cmd/commands/datasetIngestor.go
consolethinks: due to the way Go handles stuff, it was package wide. I can return a number for the current dataset, but not for the whole list. This can be added to a variable local to the ingestor function to keep track of the total (the original behaviour). Even if I did not use callbacks, I'd still need this variable.

cmd/commands/datasetIngestor.go
consolethinks: There's no reason now, I'll remove it.

cmd/commands/datasetIngestor.go
consolethinks: And global variables ("file wide") are almost never a good idea, they'll certainly cause issues in a threaded environment (eg. REST server). Normally the variable should be separate for each ingestion process, which can only be done (or the easiest way at least) if the variable is local to the function call.

datasetIngestor/getLocalFileList.go
```diff
@@ -261,22 +191,34 @@ Do you want to keep the link in dataset or skip it (D(efault)/k(eep)/s(kip) ?`,
...
 		})
 
-		if e != nil {
-			log.Fatal("Fatal error:", e)
+		if err != nil {
+			return []Datafile{}, time.Time{}, time.Time{}, "", 0, 0, fmt.Errorf("file walk returned error: %v", err)
 		}
 	}
 	// spin.Stop()
 	return fullFileArray, startTime, endTime, owner, numFiles, totalSize, err
```
minottic, line 199: I meant why not returning 
fullFileArray, startTime, endTime, owner, numFiles, totalSize, **skippedLinks, illegalFileNames**, err

cmd/commands/datasetIngestor.go
minottic: sorry, my comment was badly written. See [here](https://github.com/paulscherrerinstitute/scicat-cli/pull/104#discussion_r1731038141) what I meant

datasetIngestor/getLocalFileList.go
consolethinks: skippedLinks and illegalFilenames are both handled by the callback functions, so if counting is important, they can do it. That's how I written up this PR as well. 

Another minor reason is that having too many returns also makes the code less accessible. 

Although, while I feel like the way I set it up is pretty interesting, it's also not the easiest to understand for a beginner in Go. 

I'm against changing it, but if you're convinced, just let me know and I'll change it.

minottic: the counting can still happen from within the callbacks as long as they return it. I am fine with your implementation, I tend to avoid changes "in place". Maybe a way in between is to pass a struct to GetFiles rather than a callback and this struct has an internal counter? But otherwise, let's keep it as is!

minottic (approved):

PR #105, opened 2024-08-23
Refactor/various fixes

consolethinks:
Just a couple of commits that fix a few issues:
 - adds a few extra comments to datasetIngestor
 - adds a few extra log messages to datasetIngestor
 - "skip" variable has been renamed to skipSymlinks for better clarity
 - re-added the reset block for skipSymlinks for maintaining previous behaviour
 - fix an error message in ingestDataset

### DIFF
--- a/cmd/commands/datasetIngestor.go
+++ b/cmd/commands/datasetIngestor.go
@@ -132,7 +132,7 @@ For Windows you need instead to specify -user username:password on the command l
 			return
 		}
 
-		// check for program version
+		// === check for program version ===
 		datasetUtils.CheckForNewVersion(client, CMD, VERSION)
 		datasetUtils.CheckForServiceAvailability(client, testenvFlag, autoarchiveFlag)
 
@@ -214,7 +214,9 @@ For Windows you need instead to specify -user username:password on the command l
 		}
 		color.Set(color.FgYellow)
 		if len(foundList) > 0 {
-			fmt.Println("Warning! The following datasets have been found with the same sourceFolder: ")
+			fmt.Println("Warning! The following datasets have been found with the same sourceFolders: ")
+		} else {
+			log.Println("Finished testing for existing source folders.")
 		}
 		for _, element := range foundList {
 			fmt.Printf("  - PID: \"%s\", sourceFolder: \"%s\"\n", element.Pid, element.SourceFolder)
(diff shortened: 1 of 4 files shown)


cmd/commands/datasetIngestor.go
```diff
@@ -267,15 +268,21 @@ For Windows you need instead to specify -user username:password on the command l
...
+			// === get filelist of dataset ===
+			log.Printf("Getting filelist for \"%s\"...\n", datasetSourceFolder)
 			fullFileArray, startTime, endTime, owner, numFiles, totalSize, err :=
-				datasetIngestor.GetLocalFileList(datasetSourceFolder, datasetFileListTxt, &skip)
+				datasetIngestor.GetLocalFileList(datasetSourceFolder, datasetFileListTxt, &skipSymlinks)
 			if err != nil {
 				log.Fatalf("Can't gather the filelist of \"%s\"", datasetSourceFolder)
 			}
-
+			log.Println("Filelist collected.")
```
minottic, line 283: missing space

datasetIngestor/ingestDataset.go
```diff
@@ -176,7 +176,7 @@ func createOrigDatablocks(client *http.Client, APIServer string, fullFileArray [
 
 	if totalFiles > TOTAL_MAXFILES {
 		return fmt.Errorf(
-			"dataset exceeds (%v) the maximum number of files per dataset , which can currently be handled by the archiving system (%v)",
+			"dataset exceeds (%v) the maximum number of files per dataset , which cannot currently be handled by the archiving system (%v)",
```
minottic, line 179: I think it was correct before

minottic (commented):
this will have conflicts with #104 I think, otherwise good, minor comments

datasetIngestor/ingestDataset.go
consolethinks: I think it's a poorly written message because the original version has a double meaning: "exceeds the maximum number of files that can otherwise be handled by the archival system" (not true) and "exceeds the maximum number of files that can be handled by the archival system" (true).

I'll rewrite this again.

cmd/commands/datasetIngestor.go
consolethinks: I'll fix it

cmd/commands/datasetIngestor.go
```diff
@@ -236,8 +238,7 @@ For Windows you need instead to specify -user username:password on the command l
 		// TODO ask archive system if sourcefolder is known to them. If yes no copy needed, otherwise
 		// a destination location is defined by the archive system
 		// for now let the user decide if he needs a copy
-
-		if nocopyFlag {
+		if nocopyFlag || beamlineAccount {
```
minottic, line 241: why this?

consolethinks: Screenshot from 2024-08-26 14-34-26-mod
This is what confused me. Thinking about it now, it looks like it wants to set copy to false IF neither the copy nor nocopy are set, but at that point copy is by default false. The "checkCentralAvailability" already disables *checking* if the user is a beamline account, and since copyFlag is false by default... I think the previous version is correct. I'll restore it

consolethinks: This code is from before the refactoring stuff (v3.0.0-alpha2)

minottic (approved):

PR #107, opened 2024-08-27
Small fixes for ingestor

consolethinks:
- fix the check for changed 'allowexisitingsourcefolder' flag, 
- fix symlink handler

### DIFF
--- a/cmd/commands/datasetIngestor.go
+++ b/cmd/commands/datasetIngestor.go
@@ -223,7 +223,7 @@ For Windows you need instead to specify -user username:password on the command l
 		}
 		color.Unset()
 		if !allowExistingSourceFolder && len(foundList) > 0 {
-			if cmd.Flags().Changed("allowexistingsource") {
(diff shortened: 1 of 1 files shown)


cmd/commands/datasetIngestor.go
```diff
@@ -223,7 +223,7 @@ For Windows you need instead to specify -user username:password on the command l
 		}
 		color.Unset()
 		if !allowExistingSourceFolder && len(foundList) > 0 {
-			if cmd.Flags().Changed("allowexistingsource") {
+			if !cmd.Flags().Changed("allowexistingsource") {
```
minottic, line 226: isn't the if on top enough?

consolethinks: That's why I had to correct my code because I forgot to put the negation there ("if it was ***not*** set explicitly then ask the user")

minottic (approved):

PR #108, opened 2024-08-27
Feature/globus transfer

consolethinks:
Adds Globus as a way to transfer files:
- integrate globus go library written for the OpenEM project
- abstraction functions for SSH (rsync) and Globus transfers
- abstracted config for both transfer types
- add new command called "globusCheckTransfer" for checking transfers
- "globusCheckTransfer" supports marking datasets as archivable if the transfer succeeds

### DIFF
--- a/cmd/cliutils/globusLogin.go
+++ b/cmd/cliutils/globusLogin.go
@@ -1,0 +1,56 @@
+package cliutils
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/SwissOpenEM/globus"
+	"golang.org/x/oauth2"
+	"gopkg.in/yaml.v2"
+)
+
+type globusConfig struct {
+	ClientID              string   `yaml:"client-id"`
+	ClientSecret          string   `yaml:"client-secret,omitempty"`
+	RedirectURL           string   `yaml:"redirect-url"`
+	Scopes                []string `yaml:"scopes,omitempty"`
+	SourceCollection      string   `yaml:"source-collection"`
+	DestinationCollection string   `yaml:"dest-collection"`
+}
+
+func GlobusLogin(confPath string) (gClient globus.GlobusClient, srcCollection string, destCollection string, err error) {
+	// read in config
+	data, err := os.ReadFile(confPath)
+	if err != nil {
+		return globus.GlobusClient{}, "", "", fmt.Errorf("can't read globus config: %v", err)
+	}
+	var gConfig globusConfig
+	err = yaml.Unmarshal(data, &gConfig)
+	if err != nil {
+		return globus.GlobusClient{}, "", "", fmt.Errorf("can't unmarshal globus config: %v", err)
+	}
+
+	// config setup
+	ctx := context.Background()
+	clientConfig := globus.AuthGenerateOauthClientConfig(ctx, gConfig.ClientID, gConfig.ClientSecret, gConfig.RedirectURL, gConfig.Scopes)
+	verifier := oauth2.GenerateVerifier()
+	clientConfig.AuthCodeURL("state", oauth2.AccessTypeOffline, oauth2.S256ChallengeOption(verifier))
+
+	// redirect user to consent page to ask for permission and obtain the code
+	url := clientConfig.AuthCodeURL("state", oauth2.AccessTypeOffline, oauth2.S256ChallengeOption(verifier))
+	fmt.Printf("Visit the URL for the auth dialog: %v\n\nEnter the received code here: ", url)
+
+	// negotiate token and create client
+	var code string
+	if _, err := fmt.Scan(&code); err != nil {
+		return globus.GlobusClient{}, "", "", err
+	}
+	tok, err := clientConfig.Exchange(ctx, code, oauth2.VerifierOption(verifier))
+	if err != nil {
+		return globus.GlobusClient{}, "", "", fmt.Errorf("oauth2 exchange failed: %v", err)
+	}
+
+	// return globus client
+	return globus.HttpClientToGlobusClient(clientConfig.Client(ctx, tok)), gConfig.SourceCollection, gConfig.DestinationCollection, nil
+}
--- a/cmd/cliutils/globusLogin.go
+++ b/cmd/cliutils/globusLogin.go
@@ -16,19 +16,21 @@ type globusConfig struct {
 	RedirectURL           string   `yaml:"redirect-url"`
 	Scopes                []string `yaml:"scopes,omitempty"`
 	SourceCollection      string   `yaml:"source-collection"`
-	DestinationCollection string   `yaml:"dest-collection"`
+	SourcePrefixPath      string   `yaml:"source-prefix-path,omitempty"`
+	DestinationCollection string   `yaml:"destination-collection"`
+	DestinationPrefixPath string   `yaml:"destination-prefix-path,omitempty"`
 }
 
-func GlobusLogin(confPath string) (gClient globus.GlobusClient, srcCollection string, destCollection string, err error) {
+func GlobusLogin(confPath string) (gClient globus.GlobusClient, srcCollection string, srcPrefixPath string, destCollection string, destPrefixPath string, err error) {
 	// read in config
 	data, err := os.ReadFile(confPath)
 	if err != nil {
-		return globus.GlobusClient{}, "", "", fmt.Errorf("can't read globus config: %v", err)
+		return globus.GlobusClient{}, "", "", "", "", fmt.Errorf("can't read globus config: %v", err)
 	}
 	var gConfig globusConfig
 	err = yaml.Unmarshal(data, &gConfig)
 	if err != nil {
-		return globus.GlobusClient{}, "", "", fmt.Errorf("can't unmarshal globus config: %v", err)
+		return globus.GlobusClient{}, "", "", "", "", fmt.Errorf("can't unmarshal globus config: %v", err)
 	}
 
 	// config setup
@@ -44,13 +46,13 @@ func GlobusLogin(confPath string) (gClient globus.GlobusClient, srcCollection st
 	// negotiate token and create client
 	var code string
 	if _, err := fmt.Scan(&code); err != nil {
-		return globus.GlobusClient{}, "", "", err
+		return globus.GlobusClient{}, "", "", "", "", err
 	}
 	tok, err := clientConfig.Exchange(ctx, code, oauth2.VerifierOption(verifier))
 	if err != nil {
-		return globus.GlobusClient{}, "", "", fmt.Errorf("oauth2 exchange failed: %v", err)
+		return globus.GlobusClient{}, "", "", "", "", fmt.Errorf("oauth2 exchange failed: %v", err)
 	}
 
 	// return globus client
-	return globus.HttpClientToGlobusClient(clientConfig.Client(ctx, tok)), gConfig.SourceCollection, gConfig.DestinationCollection, nil
+	return globus.HttpClientToGlobusClient(clientConfig.Client(ctx, tok)), gConfig.SourceCollection, gConfig.SourcePrefixPath, gConfig.DestinationCollection, gConfig.DestinationPrefixPath, nil
 }
--- a/cmd/cliutils/globusLogin.go
+++ b/cmd/cliutils/globusLogin.go
@@ -10,7 +10,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-type globusConfig struct {
+type GlobusConfig struct {
 	ClientID              string   `yaml:"client-id"`
 	ClientSecret          string   `yaml:"client-secret,omitempty"`
 	RedirectURL           string   `yaml:"redirect-url"`
@@ -21,16 +21,15 @@ type globusConfig struct {
 	DestinationPrefixPath string   `yaml:"destination-prefix-path,omitempty"`
 }
 
-func GlobusLogin(confPath string) (gClient globus.GlobusClient, srcCollection string, srcPrefixPath string, destCollection string, destPrefixPath string, err error) {
+func GlobusLogin(confPath string) (gClient globus.GlobusClient, gConfig GlobusConfig, err error) {
 	// read in config
 	data, err := os.ReadFile(confPath)
 	if err != nil {
-		return globus.GlobusClient{}, "", "", "", "", fmt.Errorf("can't read globus config: %v", err)
+		return globus.GlobusClient{}, GlobusConfig{}, fmt.Errorf("can't read globus config: %v", err)
 	}
-	var gConfig globusConfig
 	err = yaml.Unmarshal(data, &gConfig)
 	if err != nil {
-		return globus.GlobusClient{}, "", "", "", "", fmt.Errorf("can't unmarshal globus config: %v", err)
+		return globus.GlobusClient{}, GlobusConfig{}, fmt.Errorf("can't unmarshal globus config: %v", err)
 	}
 
 	// config setup
@@ -46,13 +45,13 @@ func GlobusLogin(confPath string) (gClient globus.GlobusClient, srcCollection st
 	// negotiate token and create client
 	var code string
 	if _, err := fmt.Scan(&code); err != nil {
-		return globus.GlobusClient{}, "", "", "", "", err
+		return globus.GlobusClient{}, GlobusConfig{}, err
 	}
 	tok, err := clientConfig.Exchange(ctx, code, oauth2.VerifierOption(verifier))
 	if err != nil {
-		return globus.GlobusClient{}, "", "", "", "", fmt.Errorf("oauth2 exchange failed: %v", err)
+		return globus.GlobusClient{}, GlobusConfig{}, fmt.Errorf("oauth2 exchange failed: %v", err)
 	}
 
 	// return globus client
-	return globus.HttpClientToGlobusClient(clientConfig.Client(ctx, tok)), gConfig.SourceCollection, gConfig.SourcePrefixPath, gConfig.DestinationCollection, gConfig.DestinationPrefixPath, nil
+	return globus.HttpClientToGlobusClient(clientConfig.Client(ctx, tok)), gConfig, nil
 }
--- a/cmd/cliutils/globusTransfer.go
+++ b/cmd/cliutils/globusTransfer.go
@@ -9,6 +9,8 @@ func GlobusTransfer(params TransferParams) (archivable bool, err error) {
 	// === collecting params. ===
 	globusClient := params.GlobusClient
 
+	fileList := params.Filelist
+	isSymlinkList := params.IsSymlinkList
 	srcCollection := params.SrcCollection
 	dsSourceFolder := params.DatasetSourceFolder
 
@@ -20,7 +22,7 @@ func GlobusTransfer(params TransferParams) (archivable bool, err error) {
 
 	// === copying files ===
 	log.Println("Syncing files to cache server...")
-	result, err := globusClient.TransferFileList(srcCollection, dsSourceFolder, destCollection, destFolder, []string{}, []bool{})
+	result, err := globusClient.TransferFileList(srcCollection, dsSourceFolder, destCollection, destFolder, fileList, isSymlinkList, true)
 	log.Printf("The transfer result response: \n=====\n%v\n=====\n", result)
 	log.Println("Syncing files - STARTED")
 
--- a/cmd/cliutils/globusTransfer.go
+++ b/cmd/cliutils/globusTransfer.go
@@ -1,0 +1,29 @@
+package cliutils
+
+import (
+	"log"
+	"strings"
+)
+
+func GlobusTransfer(params TransferParams) (archivable bool, err error) {
+	// === collecting params. ===
+	globusClient := params.GlobusClient
+
+	srcCollection := params.SrcCollection
+	dsSourceFolder := params.DatasetSourceFolder
+
+	destCollection := params.DestCollection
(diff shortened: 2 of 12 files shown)


datasetIngestor/getLocalFileList.go
```diff
@@ -146,19 +147,20 @@ func GetLocalFileList(sourceFolder string, filelistingPath string, symlinkCallba
 			if runtime.GOOS == windows {
 				modpath = strings.Replace(path, "\\", "/", -1)
 			}
-			fileStruct := Datafile{Path: modpath, User: uidName, Group: gidName, Perm: f.Mode().String(), Size: f.Size(), Time: f.ModTime().Format(time.RFC3339)}
+			fileStruct := Datafile{Path: modpath, User: uidName, Group: gidName, Perm: f.Mode().String(), Size: f.Size(), Time: f.ModTime().Format(time.RFC3339), IsSymlink: false}
```
minottic, line 150: is this false because windows does not manage symlinks properly? If so, is this told to the users anywhere?

cmd/commands/globusCheckTransfer.go
```diff
@@ -0,0 +1,260 @@
...
+
+		globusClient, _, srcPrefixPath, _, _, err := cliutils.GlobusLogin(globusConfigPath)
+		if err != nil {
+			log.Fatalf("Couldn't create globus client: %v\n", err)
+		}
+
+		// go through each transfer task, and execute the requested operations
+		var archivableDatasetList []string
+		for _, taskId := range args {
+			task, err := globusClient.TransferGetTaskByID(taskId)
```
minottic, line 147: so get taskById returns a list of tasks? I would expect it to return one

cmd/commands/globusCheckTransfer.go
minottic, line 1: maybe this can be split into smaller funcs?

minottic (commented):
some comments, but most important we should add tests here

cmd/commands/globusCheckTransfer.go
consolethinks: What do you mean? It returns one task. I just loop over the list of tasks given as argument.

cmd/commands/datasetIngestor.go
```diff
@@ -481,12 +547,14 @@ func init() {
 	datasetIngestorCmd.Flags().Bool("noninteractive", false, "If set no questions will be asked and the default settings for all undefined flags will be assumed")
 	datasetIngestorCmd.Flags().Bool("copy", false, "Defines if files should be copied from your local system to a central server before ingest (i.e. your data is not centrally available and therefore needs to be copied ='decentral' case). copyFlag has higher priority than nocopyFlag. If neither flag is defined the tool will try to make the best guess.")
 	datasetIngestorCmd.Flags().Bool("nocopy", false, "Defines if files should *not* be copied from your local system to a central server before ingest (i.e. your data is centrally available and therefore does not need to be copied ='central' case).")
+	datasetIngestorCmd.Flags().String("transfer-type", "ssh", "Selects the transfer type to be used for transferring files. Available options: \"ssh\", \"globus\"")
```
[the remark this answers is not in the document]
consolethinks: It's literally not rsync under Windows, so I just named it based on the underlying protocol.

cmd/commands/datasetIngestor.go
```diff
@@ -239,7 +282,7 @@ For Windows you need instead to specify -user username:password on the command l
 		// a destination location is defined by the archive system
 		// for now let the user decide if he needs a copy
 
-		if nocopyFlag {
+		if nocopyFlag || beamlineAccount {
```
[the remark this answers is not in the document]
consolethinks: This PR was rebased many times, I'll fix it on Monday

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
consolethinks: This implements a CLI based login process using the functions in the `globus` package, which otherwise shouldn't be a part of a general use case library. The code in this file won't ever be used by either the REST service nor the GUI app. At most, we'll generate the redirect URL and do the authentication exchange in the same way, which totals to 6 lines of code in that file, the rest is very specific to CLI.

cmd/cliutils/globusLogin.go
```diff
@@ -0,0 +1,58 @@
...
+	if _, err := fmt.Scan(&code); err != nil {
+		return globus.GlobusClient{}, "", "", "", "", err
+	}
+	tok, err := clientConfig.Exchange(ctx, code, oauth2.VerifierOption(verifier))
+	if err != nil {
+		return globus.GlobusClient{}, "", "", "", "", fmt.Errorf("oauth2 exchange failed: %v", err)
+	}
+
+	// return globus client
+	return globus.HttpClientToGlobusClient(clientConfig.Client(ctx, tok)), gConfig.SourceCollection, gConfig.SourcePrefixPath, gConfig.DestinationCollection, gConfig.DestinationPrefixPath, nil
```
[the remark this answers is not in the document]
consolethinks: Could change that yeah, originally I didn't have the prefix path returns, and it was simpler. Not sure if it would pose a security risk keeping the client-id and secret in memory though, but that's overthinking it a little.

cmd/commands/globusCheckTransfer.go
consolethinks: I'm not sure in which way you want this to be split. 

If this function would only tell the user what tasks have finished, then they'd need to manually call a separate function that tells them which dataset the task is transferring, then a third one that just blindly marks the dataset with given datasetIds as archivable. 

Or, if we want to avoid being able to mark any datasets as archivable, the command needs to ask for transfer task id's, at which point, we have gotten to this command's complexity level again.

datasetIngestor/getLocalFileList.go
consolethinks: That's the default value, the test for symlinks is just a few lines below where it sets `fileStruct.IsSymlink = true` if the check passes.

cmd/commands/globusCheckTransfer.go
minottic: ofc, my bad, sorry. Why would people need to check more than 1 globus task at a time though? And wouldn't they want to check by jobID rather than globus task?

cmd/commands/globusCheckTransfer.go
minottic: no, I just meant internal functions to easy code readability. For example §97 to 217 could be one function, as well as line 228 to 238

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
minottic: mmm ok, I am not sure I see which parts are not relevant for the rest auth though. Isn't the `AccessTypeOffline` the only part changing?

cmd/commands/globusCheckTransfer.go
consolethinks: There's no archival job until we're certain that the files are on the PSI side, I can't use jobID for that. Unless you mean some other jobID, but I don't think there's one that fits the bill.

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
consolethinks: The config file reading shouldn't be part of the library, and anything related to stdin reading/stdout writing definitely shouldn't be there. Which is what the majority of that code is. Like I said, like 6 lines of code would be at most shared between, which would still require like 3 function calls if I put them into library functions. In my opinion, the returns are diminishing trying to jam those 6 lines into functions.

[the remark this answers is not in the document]
minottic: but so the user will everytime need to set their globus account credentials? And this feature is only provided to people with a proper globus account? Should there be then a fallback to ssh in cases where users do not have a valid globus account (basically when the auth fails)?

cmd/commands/globusCheckTransfer.go
minottic: ah right, I always get confused by that. So the user needs to get the list of taskIds somehow? Maybe it's worth printing the taskIds when the (globus) ingestion exists, so they can easily see them, without logging into globus. I also have the impression that the job should be created before the transfer, but not a strong opinion if users see the tasks list printed by the ingestor

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
sbliven: It seems reasonable to keep the oauth2 negotiation outside the globus library.

Maybe it makes sense to fall back to ssh eventually, but I could also see the argument for just failing with an error if globus is requested but fails. I do think we should implement session caching somehow to avoid re-authenticating between runs, but I think this can be added after this PR.

cmd/cliutils/globusTransfer.go
```diff
@@ -0,0 +1,42 @@
...
+
+	destCollection := params.DestCollection
+	destPrefixPath := params.DestPrefixPath
+	datasetId := params.DatasetId
+
+	archivable = false // the dataset is never archivable after a globus transfer request immediately
+	destFolder := destPrefixPath + "/archive/" + strings.Split(datasetId, "/")[1] + dsSourceFolder
+
+	for i := range fileList {
+		fileList[i] = srcPrefixPath + "/" + fileList[i]
```
sbliven, line 26: This doesn't seem platform-independent. Shouldn't it use `filepath.Join` and `filepath.ToSlash` to convert to valid linux paths?

cmd/commands/datasetIngestor.go
```diff
@@ -68,14 +68,57 @@ For Windows you need instead to specify -user username:password on the command l
...
+			log.Fatalln(err)
+		}
+
+		var transferFiles func(params cliutils.TransferParams) (archivable bool, err error)
+
+		// globus specific vars (if needed)
+		var globusClient globus.GlobusClient
+		var srcCollection, srcPrefixPath, destCollection, destPrefixPath string
+
+		switch transferType {
```
sbliven, line 96: Shouldn't this configuration be skipped if copying is not required?

cmd/commands/datasetIngestor.go
```diff
@@ -68,14 +68,57 @@ For Windows you need instead to specify -user username:password on the command l
...
+				globusConfigPath = filepath.Join(filepath.Dir(execPath), "globus.yaml")
+			}
+
+			globusClient, srcCollection, srcPrefixPath, destCollection, destPrefixPath, err = cliutils.GlobusLogin(globusConfigPath)
+			if err != nil {
+				log.Fatalln("couldn't create globus client:", err)
+			}
+
+			if autoarchiveFlag {
+				log.Println("Cannot autoarchive when transferring via Globus due to the transfer happening asynchronously. Use the \"globusCheckTransfer\" command to archive them")
```
sbliven, line 118: Maybe should be Fatal to avoid user mistakes?

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
consolethinks: One can save the received token and use it as long as it is valid. I just didn't bother setting that up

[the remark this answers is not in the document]
consolethinks: And when I was talking about the "user should see how they want to store the config" I meant the library user (programmer), not the end user

cmd/commands/globusCheckTransfer.go
sbliven: This would require us adding some sort of `hold` status to the job lifecycle so that scicat/nodered can postpone the creation actions until later. I think it increases the complexity.

cmd/commands/globusCheckTransfer.go
```diff
@@ -0,0 +1,260 @@
...
+
+	"github.com/fatih/color"
+	"github.com/paulscherrerinstitute/scicat/cmd/cliutils"
+	"github.com/paulscherrerinstitute/scicat/datasetIngestor"
+	"github.com/paulscherrerinstitute/scicat/datasetUtils"
+	"github.com/spf13/cobra"
+)
+
+var globusCheckTransfer = &cobra.Command{
+	Use:   "globusCheckTransfer [options] (transfer_task_id transfer_task_id ...)",
```
sbliven, line 21: So users need to remember the globus task id? It feels more convenient to just use the PID.

datasetIngestor/getLocalFileList.go
```diff
@@ -146,19 +147,20 @@ func GetLocalFileList(sourceFolder string, filelistingPath string, symlinkCallba
 			if runtime.GOOS == windows {
 				modpath = strings.Replace(path, "\\", "/", -1)
```
sbliven, line 148: Old code, but filepath.ToSlash would be clearer

sbliven (approved):

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
minottic: so whose are the clientId and secretId? I thought they were the end-user's

cmd/cliutils/globusTransfer.go
consolethinks: Globus always uses forward slashes, so I actually went out of my way to do it this way. It also means that the filelist produced by a previous step might not work properly when run under Windows. That could be handled by a separate PR I feel like or I could add a rudamentary "if os == windows, replace '\' with '/'"

consolethinks: Source: https://docs.globus.org/globus-connect-personal/install/windows/ (the first NOTE section)

cmd/commands/datasetIngestor.go
consolethinks: Good idea I'll change it.

cmd/commands/globusCheckTransfer.go
consolethinks: The task ID is listed by datasetIngestor when ingesting datasets with the globus transfer option. 

If there's a way to add as a tag the globus task id associated with the dataset in SciCat, that would work as well. The tag could be removed by this command afterwards. 

If we are certain that PID's will never exceed 128 characters, another option would be using the "label" attribute of Globus transfers where I can put in the PID of datasets, then use a filter by label when searching for transfers. https://docs.globus.org/api/transfer/task_submit/
image

cmd/commands/globusCheckTransfer.go
consolethinks: The task ID is printed out by GlobusTransfer(params) in cmd/cliutils/globusTransfer.go.

datasetIngestor/getLocalFileList.go
consolethinks: this code snippet converts from backslash to forward slash when running Windows. filepath.ToSlash would only do forward slash to backslash under Windows.

cmd/commands/datasetIngestor.go
consolethinks: I could add an if block around this. My thinking was that I handle the transfer selection in advance so that the rest of the code can operate in the same way with minimal changes.

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
consolethinks: the client secret is optional, the client id is there to authorize the thick client to interact with Globus. Normally, an administrator needs to create a new project, then add thick client whose client id is the one that should be used with the thick client application. You can do this [here](https://app.globus.org/settings/developers). The client id and (optional) secret could be made part of the executable as well, as they should not change at all normally.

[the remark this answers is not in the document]
minottic: ah right, ofc. since we might want to be cautious with clientIds and clientSecret sniffing (even if they are useless without valid globus credentials), I'd probably supply the clientId only during GO build, so at least they are not in github. 

Another option would be getting the oauth2 token through an auth server that has the clientId and clientSecret, but I am unsure this is feasible with this flow.

And what's the redirectURL in this case?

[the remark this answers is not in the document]
consolethinks: The one that you have given during app registration. 

You can read this to know more:
https://docs.globus.org/api/auth/developer-guide/#register-app

cmd/commands/globusCheckTransfer.go
consolethinks: Actually are you sure you want me to make changes like that for this PR? These kinds of changes could be its own PR.

minottic: as you wish, since this is new code I think they could also live here, but as you want

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
sbliven: If you use the [PKCE](https://docs.globus.org/api/auth/developer-guide/#pkce) flow then you don't need to register beforehand. This is preferred for apps that get distributed since there is no secret.

cmd/cliutils/globusTransfer.go
sbliven: I think that's what filepath.ToSlash does, except it would also apply to srcPrefixPath.

cmd/commands/globusCheckTransfer.go
consolethinks: Oh I thought this was for the datasetIngestor! I'll check what I can do then

cmd/commands/datasetIngestor.go
```diff
@@ -91,32 +91,33 @@ For Windows you need instead to specify -user username:password on the command l
...
-			var globusConfigPath string
-			if cmd.Flags().Lookup("globus-cfg").Changed {
-				globusConfigPath = globusCfgFlag
-			} else {
-				execPath, err := os.Executable()
-				if err != nil {
-					log.Fatalln("can't find executable path:", err)
+		var gConfig cliutils.GlobusConfig
+
+		if copyFlag {
```
sbliven, line 96: This doesn't quite work since we don't know if the data will actually be copied until CheckDataCentrallyAvailableSsh. Maybe it does need to be done in all cases.

I don't like how the logic about whether or not to transfer gets interleaved into all the other functions. Maybe this is something that should be refactored later.

cmd/commands/datasetIngestor.go
sbliven: This might have been a bad suggestion (see below)

cmd/commands/globusCheckTransfer.go
sbliven: I like the idea of using label to store the PID.

datasetIngestor/getLocalFileList.go
sbliven: > [ToSlash](https://pkg.go.dev/path/filepath#ToSlash) returns the result of replacing each separator character in path with a slash ('/') character

sbliven: This doesn't really matter since it only runs on windows anyways

cmd/commands/globusCheckTransfer.go
consolethinks: okay, so there are a few options: 

1. create a new package specifically for globusCheckTransfer in order to separate its functions from the rest of the cmd package. The functions will have 6-8 (eg. for handling a task, we need globusClient, globus config (for source and dest path prefixes), client, APIServer string, user (for scicat connection), taskId, markArchivable bool flag, skipDestPathCheck flag, dryRun flag)

2. write the functions in the same package as the commands ("cmd" package), which means I'll have to give it a highly specific name (eg. globusCheckTransferHandleJob) due to the function name being public to all files within the same package. Also would have similarly long parameter lists.

3. Write the function as an internal function within the command. This would hide the function declaration from other files, so it doesn't need a specific name. And since it's declared within the command's function block, it can access the parameters it needs without explicitly passing them as parameters.

I presonally would keep it as is, as any solution would highly diverge from the convention of the other comments in the current state

cmd/commands/datasetIngestor.go
sbliven: Sorry for the additional effort.

cmd/commands/globusCheckTransfer.go
minottic: either 2 or why not functions inside `globusCheckTransfer.go` (maybe that's what you meant with 3)? They will also be easier to test if dedicated functions

cmd/cliutils/globusTransfer.go
consolethinks: Yeah, I checked the exact descriptions of "fromSlash" and "toSlash" and that seems to be the case.

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
consolethinks: I don't exaclty know what they mean by unauthenticated client, but the example URL given still contains a client_id:
`
https://auth.globus.org/v2/oauth2/authorize?code_challenge=rennw4QyOe3rIlZq-qTh2gL34pYEC_5JXKSKRhc5lVc&code_challenge_method=S256&state=_default&redirect_uri=https%3A%2F%2Fauth.globus.org%2Fv2%2Fweb%2Fauth-code&response_type=code&client_id=asdf&scope=openid+profile+email&access_type=online
`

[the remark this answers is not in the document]
sbliven: I think the client id is public. The advantage of PKCE is that there's no secret id.

cmd/commands/globusCheckTransfer.go
consolethinks: I'll do this in a later PR, it requires significant rework that involves the Globus library as well.

cmd/commands/globusCheckTransfer.go
consolethinks: I have added two functions in globusCheck Transfer in the latest commits, hopefully this is enough for better code comprehensibility as there aren't many other places where large blocks of code would have a sensible refactor to functions.

cmd/cliutils/globusLogin.go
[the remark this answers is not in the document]
consolethinks: and that's what I am doing right now then, as I'm not using any secrets

PR #109, opened 2024-08-27
Update golangci-lint

consolethinks:
Update golangci-lint from v1.50.1 to v1.60. This is needed to ensure that it works with the latest go versions and corresponding packages.


minottic (approved):
maybe this is an opportunity to set some dependency bot (e.g. renovate)

PR #110, opened 2024-08-27
OpenEM: Draft of how an ingestion flow could look like

phwissmann:


datasetIngestor/ingestDataset.go
```diff
@@ -114,7 +114,8 @@ func createDataset(client *http.Client, APIServer string, metaDataMap map[string
 func getEndpoint(dstype string) (string, error) {
 	switch dstype {
 	case "raw":
-		return "/RawDatasets", nil
+		// return "/RawDatasets", nil
+		return "/Datasets", nil
```
sbliven, line 118: This is a v3 to v4 difference. V3 had separate endpoints for all classes, whereas V4 just has the main `/dataset` endpoint.

There is a `/Datasets` in v3. Maybe that could be used everywhere and still be backwards compatible?

datasetIngestor/ingestDataset.go
```diff
@@ -114,7 +114,8 @@ func createDataset(client *http.Client, APIServer string, metaDataMap map[string
...
 	case "derived":
 		return "/DerivedDatasets", nil
```
sbliven, line 120: In V4 derived datasets should also use the `/datasets` endpoint.

openem-ingestor/core/datasetingestor.go
```diff
@@ -0,0 +1,109 @@
...
+	task IngestionTask,
+) (string, error) {
+
+	var http_client = &http.Client{
+		Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}},
+		Timeout:   120 * time.Second}
+
+	SCIAT_API_URL := task.ScicatUrl
+
+	const TAPECOPIES = 2 // dummy value, unused
```
sbliven, line 28: default in the CLI is 1. It should match if this gets implemented

openem-ingestor/core/datasetingestor.go
```diff
@@ -0,0 +1,109 @@
...
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/paulscherrerinstitute/scicat/datasetIngestor"
+)
+
+func IngestDataset(
```
sbliven, line 16: There seems to be a lot of bits in `IngestDataset` that could be moved into the datasetIngestor package. I feel like core/datasetingestor should focus on executing the task. The fact that it has to run lots of different functions (CheckMetadata, UpdateMetaData, Upload*, MarkFIlesReady) seems to indicate a poorly designed base library. Should we fix that rather than duplicating code here and in `cmd`?

openem-ingestor/core/taskqueue.go
```diff
@@ -0,0 +1,161 @@
...
+	resultChannel        chan TaskResult    // The result of the upload is put into this channel
+	AppContext           context.Context
+}
+
+type TransferMethods int
+
+const (
+	TransferS3 TransferMethods = iota + 1
+	TransferGlobus
+)
```
sbliven, line 26: Need to coordinate this with #108

openem-ingestor/core/utils.go
```diff
@@ -0,0 +1,213 @@
...
+	"strings"
+	"time"
+
+	"github.com/fatih/color"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	"github.com/wailsapp/wails/v2/pkg/runtime"
+)
+
+func createLocalSymlinkCallbackForFileLister(skipSymlinks *string, skippedLinks *uint) func(symlinkPath string, sourceFolder string) (bool, error) {
```
sbliven, line 20: This function seems like it should be in the `datasetUtils` package too.

sbliven (commented):
A lot of the `datasetIngestor` changes are for V4 compatibility. These should be split into another PR. The ingestor should also detect the version and switch APIs as needed.

Some of my comments reflect the ongoing discussion about which repo various ingestor code should live in. I'll try to update comments after that discussion finalizes.

phwissmann:
Let me start a general discussion here since @sbliven  started to review this:

This PR was illustrative to see what is needed to make the minimum work. This is not intended to go into this repo anymore. However, it shows the following:

1. Changes to the scicat interactions are needed in order to be compatible to backend v4
2. The "dataset ingestion flow" in the cli considers a lot of non-relevant cases and several things could be simplified for the OpenEm Ingestor

Proposals:
1. create `cli-next` branch to add any changes for backend v4 (with or without v3 compatibility, tbd). Following a previous discussion, there is already a branch. I had to change the name of the module in order to align with the repo; go trips up if the names differ
2. Two options:
  a. Cleanup and abstract the  function here https://github.com/paulscherrerinstitute/scicat-cli/blob/main/cmd/commands/datasetIngestor.go#L107 and move it into the library
  b. Create a new, similar function in the OpenEm/Ingestor repo that just uses the current library functions.

In my opinion, 2b) is the easier option:
- most functionality is in the library and the duplication is relatively high-level
- the logic is application specific; the cli-tools cover more use cases (also PSI specific ones). 2a) would couple these with the OpenEm ones
 - abstracting the implementation in cli-tools would be a lot of work for only the benefit of little additional code reuse but with strong coupling to the cli-tools

sbliven:
Ok, you make good arguments for the current division of functionality.

phwissmann:
Replaced by https://github.com/SwissOpenEM/Ingestor/pull/10

PR #113, opened 2024-09-13
add major version to module path, sync module name with repo's

consolethinks:
## Description
According to [official Go guidelines](https://go.dev/doc/modules/major-version), we need to add the major version to the module path in order for people to be able to include the package as a dependency. They also recommend to create separate branches for major releases, but that's optional.


consolethinks:
As an additional note, here's what happens if I try to include v3.0,0 in a different module without the module name change:
image

sbliven (approved):
It's good to follow the golang best practices.

Looking at [managing source](https://go.dev/doc/modules/managing-source) it looks like the directory structure should match the full module name. So should it be moved to eg `github.com/paulscherrerinstitute/scicat-cli/cmd`?

consolethinks:
> It's good to follow the golang best practices.
> 
> Looking at [managing source](https://go.dev/doc/modules/managing-source) it looks like the directory structure should match the full module name. So should it be moved to eg `github.com/paulscherrerinstitute/scicat-cli/cmd`?

@sbliven 
The module encompasses the cmd, datasetUtils, datasetIngestor packages, so no change is needed because of that. However, since the module name is now scicat-cli/v3, we could move all packages into a `v3` subfolder if you think that works better. 

Personally, I prefer the solution of creating a new `v3` branch, and then we could revert main to a pre-v3 state. However, main will still be technically non conformant as its module name does not include `/v2` at the end, nor does it have its files in a `v2` subfolder, nor does it use a `v2` branch.

References on go best practices regarding major version bumps:
https://stackoverflow.com/questions/70058951/project-structure-with-v2-go-modules
https://go.dev/wiki/Modules#releasing-modules-v2-or-higher

PR #116, opened 2025-01-14
fix GetUserInfoFromToken for v4 of backend

consolethinks:
It updates the GetUserInfoFromToken function to properly interact with scicat backend v4.

### DIFF
--- a/datasetUtils/getUserInfoFromToken.go
+++ b/datasetUtils/getUserInfoFromToken.go
@@ -54,12 +54,12 @@ func GetUserInfoFromToken(client *http.Client, APIServer string, token string) (
 	if err != nil {
 		return map[string]string{}, []string{}, err
 	}
-	req2, err := http.NewRequest("GET", APIServer+"/users/"+url.QueryEscape(newUserInfo.Id)+"/userIdentity", nil)
+	filterString := url.QueryEscape(fmt.Sprintf("{\"where\":{\"userId\":\"%s\"}}", newUserInfo.Id))
+	req2, err := http.NewRequest("GET", APIServer+"/useridentities/findOne?filter="+filterString, nil)
 	if err != nil {
 		return map[string]string{}, []string{}, err
 	}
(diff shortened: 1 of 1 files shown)


datasetUtils/getUserInfoFromToken.go
```diff
@@ -54,12 +54,12 @@ func GetUserInfoFromToken(client *http.Client, APIServer string, token string) (
 	if err != nil {
 		return map[string]string{}, []string{}, err
 	}
-	req2, err := http.NewRequest("GET", APIServer+"/users/"+url.QueryEscape(newUserInfo.Id)+"/userIdentity", nil)
+	filterString := url.QueryEscape(fmt.Sprintf("{\"where\":{\"userId\":\"%s\"}}", newUserInfo.Id))
```
sbliven, line 57: The docs say
>    {
>      "where?": {
>        "profile.email": "this_email@some.site"
>      }
>    }
>  This last version is deprecated and will be discontinued as soon as the FE is updated.
>  It has been maintanined for backward compatibility.

Looks like it shouldn't have the 'where' clause.

datasetUtils/getUserInfoFromToken.go
```diff
@@ -54,12 +54,12 @@ func GetUserInfoFromToken(client *http.Client, APIServer string, token string) (
 	if err != nil {
 		return map[string]string{}, []string{}, err
 	}
-	req2, err := http.NewRequest("GET", APIServer+"/users/"+url.QueryEscape(newUserInfo.Id)+"/userIdentity", nil)
```
sbliven, line 57: This endpoint seems to still exist (at least in release-jobs). What's the issue here?

datasetUtils/getUserInfoFromToken.go
consolethinks: I copied this solution from the frontend I think, but I'll recheck with the original method for v3.

consolethinks:
This change is not necessary anymore

PR #117, opened 2025-01-27
update userpass login method for v4 backend

consolethinks:
It fixes/updates the login method for v4 backend, as the new final endpoint for it is `/auth/login` instead of `/users/login`

### DIFF
--- a/cmd/commands/authenticate.go
+++ b/cmd/commands/authenticate.go
@@ -53,8 +53,10 @@ func authenticate(authenticator Authenticator, httpClient *http.Client, apiServe
 		var user, pass string
 		uSplit := strings.Split(userpass, ":")
 		if len(uSplit) > 1 {
+			user = uSplit[0]
 			pass = uSplit[1]
 		} else {
+			user = userpass
 			fmt.Print("Password: ")
 			pw, err := term.ReadPassword(int(syscall.Stdin))
 			if err != nil {
--- a/datasetUtils/authenticateUser.go
+++ b/datasetUtils/authenticateUser.go
@@ -26,7 +26,7 @@ func AuthenticateUser(client *http.Client, APIServer string, username string, pa
 
 	// try functional accounts first
 
-	req, err := http.NewRequest("POST", APIServer+"/Users/login", bytes.NewBuffer(cred))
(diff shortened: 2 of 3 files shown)


datasetUtils/authenticateUser.go
minottic, line 1: this file has no tests at all, is it ok?

minottic (commented):
changes look good, thanks! some comments on tests

datasetUtils/authenticateUser.go
consolethinks: I can write one for it, but I feel like that's outside the scope of this PR.

cmd/commands/authenticate.go
```diff
@@ -53,8 +53,10 @@ func authenticate(authenticator Authenticator, httpClient *http.Client, apiServe
 		var user, pass string
 		uSplit := strings.Split(userpass, ":")
 		if len(uSplit) > 1 {
+			user = uSplit[0]
```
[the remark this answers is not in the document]
consolethinks:
```
		{
			name:     "Test with empty token and non-empty userpass",
			token:    "",
			userpass: "testuser:testpass",
			wantUser: map[string]string{
				"username": "testuser",
				"password": "testpass",
			},
			wantGroup: []string{"group1", "group2"},
		},
```
this test case should be the test for it, which is already part of this, but the test framework itself doesn't test behaviour properly. I'll try fixing this.

datasetUtils/authenticateUser.go
```diff
@@ -36,7 +36,8 @@ func AuthenticateUser(client *http.Client, APIServer string, username string, pa
 		return map[string]string{}, []string{}, err
 	}
 	defer resp.Body.Close()
-	if resp.StatusCode == 200 {
+
+	if resp.StatusCode == 201 {
```
sbliven, line 40: Should it accept 2xx?

sbliven (approved):

cmd/commands/authenticate.go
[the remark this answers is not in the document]
minottic: thanks!

minottic (approved):

consolethinks:
I'm going to drop this PR as I decided to rewrite the login function instead.

PR #118, opened 2025-01-29
Refactor/login rewrite

consolethinks:
This commit rewrites the login function to make it easier to read, and move PSI specific functionality to the CLI only portion of the code (KerberOS automatic user creation).


datasetUtils/authenticateUser.go
```diff
@@ -36,93 +56,57 @@ func AuthenticateUser(client *http.Client, APIServer string, username string, pa
...
-	} else {
-		// then try normal user account
-		req, err = http.NewRequest("POST", strings.Replace(APIServer, "api/v3", "auth/msad", 1), bytes.NewBuffer(cred))
-		if err != nil {
-			return map[string]string{}, []string{}, err
-		}
-		req.Header.Set("Content-Type", "application/json")
-		resp, err = client.Do(req)
+
+	if resp.StatusCode != 201 {
```
sbliven, line 60: Is the 200->201 a v4 change?

sbliven (approved):
This looks good. Some things like kinit will need to be tested in the PSI environment, but the unit tests already look good.

datasetUtils/authenticateUser.go
consolethinks: I think so, I was receiving this error code when using the v4 backend

PR #119, opened 2025-03-17
Feature/OIDC login

consolethinks:
This PR adds support for OIDC login using the additional frontend feature of scicat-backend-next. It does this by calling the backend's /auth/oidc endpoint with the "CLI" frontend client id. This means that the scicat backend has to be configured to use OIDC and with the following env vars:
 - `OIDC_FRONTEND_CLIENTS='CLI'` 
 - `OIDC_CLI_SUCCESS_URL="http://localhost:8080/success"`


sbliven (approved):
Looks good. You should rebase it after #118 merges since there's some overlap.

Note that this feature requires scicat-backend-next [4.13.0](https://github.com/SciCatProject/scicat-backend-next/releases/tag/v4.13.0)

sbliven:
This is a nice new feature. We should keep a CHANGELOG.

PR #122, opened 2025-04-24
Expand ~ to the user home in paths

sbliven:
Expand ~ in paths to the user home


consolethinks:
Normally, shouldn't `~` (tilde) be expanded into the homefolder path by the shell itself? @sbliven

PR #130, opened 2025-06-19
update createJob dto for upstream changes

consolethinks:
Moves the dataset list to top level, as that seems to be required to work with the latest version of scicat-backend-next.

### DIFF
--- a/datasetUtils/createJob.go
+++ b/datasetUtils/createJob.go
@@ -41,14 +41,14 @@ func CreateArchivalJob(client *http.Client, APIServer string, user map[string]st
 	}
 
 	type jobParamsStruct struct {
-		TapeCopies  string          `json:"tapeCopies"`
-		Username    string          `json:"username"`
-		DatasetList []datasetStruct `json:"datasetList"`
+		TapeCopies string `json:"tapeCopies"`
+		Username   string `json:"username"`
 	}
 
 	type createJobDto struct {
 		JobType      string          `json:"type"`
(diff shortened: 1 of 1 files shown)


sbliven (changes requested):
Moving datasetList is only one of the differences between CreateJobDto and CreateJobDtoV3. Please make the other changes:

Screenshot 2025-06-20 at 10 20 00

@despadam Should he also set `jobParams.username` or other undocumented properties?

datasetUtils/createJob_test.go
```diff
@@ -156,30 +156,30 @@ func TestCreateJob(t *testing.T) {
...
-						"ownerUser": "testuser",
-						"ownerGroup": "group1"
+					    {
+					      "pid": "dataset2",
+					      "files": []
+					    }
+					  ],
+					  "type": "archive",
+					  "ownerUser": "testuser",
+					  "ownerGroup": "group1",
```
despadam, line 177: These are not top-level fields in v3

despadam: If you want to provide `ownerGroup` it should be inside `jobParams`, not a top-level field.

Also `jobParams.username` corresponds to `ownerUser`, so `ownerUser` should not be provided as a top-level field.

consolethinks: I was looking at the version of scicat be that was in scicatlive. I can make these changes, but then I guess I should configure scicatlive to use the latest master branch in order to be able to test this ?

despadam (approved):

PR #131, opened 2025-06-19
use POST /api/v3/datasets for all dataset types

### LINKED ISSUE #128: Remove RawDataset/DerivedDataset endpoints
If https://github.com/SciCatProject/scicat-backend-next/issues/1966 is not implemented then the POST `/api/v3/RawDataset` and `/api/v3/DerivedDataset` endpoints will not be available. The CLI should instead use `/api/v3/Dataset` for all datasets. This might break compatibility to backend-v3 unfortunately, but I think it is necessary.

consolethinks:
As described in #128, this PR makes a change to CreateDatasetEntry function, so that it will call `POST /api/v3/Datasets` for all dataset types in order to create datasets.

This closes #128

### DIFF
--- a/datasetIngestor/createDatasetEntry_test.go
+++ b/datasetIngestor/createDatasetEntry_test.go
@@ -37,15 +37,7 @@ func TestCreateDatasetEntry(t *testing.T) {
 	}
 
 	// Check the URL based on the type field
-	expectedPath := ""
-	switch metaDataMap["type"].(string) {
-	case "raw":
-		expectedPath = "/RawDatasets"
-	case "derived":
-		expectedPath = "/DerivedDatasets"
-	case "base":
-		expectedPath = "/Datasets"
-	}
+	expectedPath := "/Datasets"
 
 	// Check the URL
 	if capturedPath != expectedPath {


datasetUtils/createJob_test.go
```diff
@@ -156,25 +156,25 @@ func TestCreateJob(t *testing.T) {
...
-						"ownerUser": "testuser",
-						"ownerGroup": "group1"
+					    {
+					      "pid": "dataset2",
+					      "files": []
+					    }
+					  ],
+					  "type": "archive",
+					  "ownerUser": "testuser",
+					  "ownerGroup": "group1"
```
despadam, line 177: If this is using the api/v3 endpoint shouldn't the dto also be updated here?

datasetUtils/createJob.go
```diff
@@ -41,14 +41,14 @@ func CreateArchivalJob(client *http.Client, APIServer string, user map[string]st
...
+		Username   string `json:"username"`
 	}
 
 	type createJobDto struct {
 		JobType      string          `json:"type"`
 		JobParams    jobParamsStruct `json:"jobParams"`
+		DatasetList  []datasetStruct `json:"datasetList"`
 		OwnerUser    string          `json:"ownerUser"`
 		OwnerGroup   string          `json:"ownerGroup"`
 		ContactEmail string          `json:"contactEmail"`
```
despadam, line 54: This dto does not correspond with api/v3

despadam (commented):
I think there is an overlap between this PR and https://github.com/paulscherrerinstitute/scicat-cli/pull/130.

Why is this PR changing things for Jobs since this has already been addressed in https://github.com/paulscherrerinstitute/scicat-cli/pull/130? I was expecting only change to the dataset endpoint here

datasetUtils/createJob.go
despadam: Yes, that would be better so that only the changes relevant to /api/v3/datasets are here

datasetUtils/createJob_test.go
consolethinks: it was rebased to an older version of the other PR, it should be fixed now

sbliven (approved):

PR #162, opened 2025-11-13
Print dataset ID on retrieve

minottic:


minottic:
see test [here](https://github.com/paulscherrerinstitute/scicat-cli/blob/add_dataset_id/datasetUtils/getDatasetDetails_test.go#L138)

PR #249, opened 2026-06-18
fix: increment counter on job creation

minottic:


datasetUtils/createJob_test.go
```diff
@@ -206,3 +206,168 @@ func TestCreateJob(t *testing.T) {
...
+}
+
+func contains(slice []string, item string) bool {
+	for _, v := range slice {
+		if v == item {
+			return true
+		}
+	}
+	return false
+}
```
fpotier: stdlib has `slices.Contains`

fpotier (approved):

datasetUtils/createJob_test.go
minottic: thanks, fixed

PR #265, opened 2026-07-13
feat: add completeIngest command

minottic:
This pr is an easier implementation of a dedicated completeIngest command available to archiveManager only and that can take one single datasetId as input, together with the token and the env to run against.

It will create origs if missing starting from a datasetId.

It introduces an `orchestrator` folder where to put logic executed by the cobra calls but which needs to be tested or benefits from single responsability or logical splits

### DIFF
--- a/cmd/commands/completeIngest.go
+++ b/cmd/commands/completeIngest.go
@@ -67,11 +67,10 @@ For further help see "` + cliutils.MANUAL + `"`,
 			return
 		}
 
-		if len(args) != 1 {
-			log.Println("invalid number of args")
-			return
+		pid, err := orchestrator.ExtractPidFromArgs(args)
+		if err != nil {
+			log.Fatal(err)
 		}
-		pid := args[0]
 
 		// === check for program version ===
 		datasetUtils.CheckForNewVersion(client, CMD, VERSION)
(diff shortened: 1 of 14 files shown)


cmd/commands/completeIngest.go
```diff
@@ -0,0 +1,108 @@
...
+			Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: false}},
+			Timeout:   120 * time.Second}
+
+		const CMD = "completeIngest"
+
+		// pass parameters
+		envConfig := cliutils.InputEnvironmentConfig{
+			TestenvFlag:  cliutils.GetCobraBoolFlag(cmd, "testenv"),
+			DevenvFlag:   cliutils.GetCobraBoolFlag(cmd, "devenv"),
+			LocalenvFlag: cliutils.GetCobraBoolFlag(cmd, "localenv"),
```
omkar-ethz: `localenv` is not added in the init, so using it gives
`Error: unknown flag: --localenv`
Maybe it's not needed?

omkar-ethz (approved):
Minor comment - LGTM!

cmd/commands/completeIngest.go
minottic: thanks, removed!

PR #266, opened 2026-07-14
fix: support archive of datasetIds without owner

minottic:
This enables using datasetArchiver with a list of datasets without needing to pass an ownerGroup that it's then inferred from user's groups. This enables submission of a single job with a mixed datasets. This is actually a fix rather than a fit, since the command was intended to work this way already, but a regression was introduced when moving to cobra

as for #265, testable code is moved to `orchestrator` folder and a bigger refactor would include moving all logic run by the datasetArchiver cobra command to the orchestrator, but this is left to a later PR

### DIFF
--- a/cmd/commands/datasetArchiver.go
+++ b/cmd/commands/datasetArchiver.go
@@ -104,11 +104,11 @@ For further help see "` + cliutils.MANUAL + `"`,
 			log.Fatal(err)
 		}
 
-		ownerGroupList, err := orchestrator.ResolveOwnerGroupList(ownerGroup, accessGroups)
+		resolvedOwnerGroup, err := orchestrator.ResolveOwnerGroup(ownerGroup, accessGroups)
 		if err != nil {
 			log.Fatal(err)
 		}
-		archivableDatasets, err := orchestrator.ResolveArchivableDatasets(client, APIServer, user["accessToken"], ownerGroupList, inputdatasetList)
+		archivableDatasets, err := orchestrator.ResolveArchivableDatasets(client, APIServer, user["accessToken"], resolvedOwnerGroup, inputdatasetList)
 		if err != nil {
 			log.Fatal(err)
 		}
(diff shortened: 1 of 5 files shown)


orchestrator/datasetArchiver.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/paulscherrerinstitute/scicat-cli/v3/datasetUtils"
+)
+
+/*
+ResolveArchivableDatasets returns the list of dataset PIDs to submit for archiving: the archivable
+datasets of ownerGroupList, optionally narrowed down to inputDatasetList. ownerGroupList is
```
omkar-ethz: I think the docstring is a little misleading. It seems to suggest that the returned list is the intersection of "archivable datasets of ownerGroupList" and the "inputDatasetList".

But if a non empty `inputDatasetList` is supplied, then ownerGroupList is completely ignored in `GetArchivableDatasets`.

We filter the inputDatasetList based on user's accessToken, which might include dataset's with other ownerGroups than the provided `ownerGroup`.

I would change it to something like:
If inputDatasetList is non-empty, returns the archivable datasets matching those PIDs that the user's accessToken can access (ownerGroupList is ignored). Otherwise, returns the archivable datasets belonging to ownerGroupList

cmd/commands/datasetArchiver.go
```diff
@@ -19,10 +20,11 @@ var datasetArchiverCmd = &cobra.Command{
 	Short: "Archives all datasets in state datasetCreated from a given ownerGroup",
 	Long: `Tool to archive datasets to the data catalog.
 
-You must choose either an ownerGroup, in which case all archivable datasets
-of this ownerGroup not yet archived will be archived.
-Or you choose a (list of) datasetIds, in which case all archivable datasets
-of this list not yet archived will be archived. 
+The ownerGroup is optional. If given, all archivable datasets of this
+ownerGroup not yet archived will be archived, optionally narrowed down to
+the (list of) datasetIds passed as arguments.
```
omkar-ethz: as commented below, i think the "narrowed down" might be confusing

omkar-ethz (approved):
some comments about the docstrings. LGTM!

cmd/commands/datasetArchiver.go
minottic: updated

PR #267, opened 2026-07-15
feat: add remote-file flag for remote ingestion

minottic:
This PR adds a `remote-files` flag which enables to run the ingestion with files not available where the cli runs. 

It follows the orchestrator folder pattern for testability and it's easier to review when reading one commit at a time

### DIFF
--- a/orchestrator/datasetIngestor.go
+++ b/orchestrator/datasetIngestor.go
@@ -4,9 +4,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"log"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/paulscherrerinstitute/scicat-cli/v3/datasetIngestor"
@@ -20,14 +20,14 @@ import (
 // must be skipped (not fatal, no os.Exit); anything else is a hard failure gathering the local
 // file list. emptyDatasets/tooLargeDatasets are incremented to match whichever of those two
 // errors is returned.
-func PrepareDataset(client *http.Client, APIServer string, user map[string]string,
+func PrepareDatasetAndUpdateCounts(client *http.Client, APIServer string, user map[string]string,
 	originalMap map[string]string, metaDataMap map[string]interface{}, tapecopies int,
 	datasetSourceFolder string, datasetFileListTxt string,
 	symlinkCallback func(symlinkPath string, sourceFolder string) (bool, error),
 	filenameCheckCallback func(filepath string) bool,
 	emptyDatasets *int, tooLargeDatasets *int) (fullFileArray []datasetIngestor.Datafile, err error) {
-	fullFileArray, startTime, endTime, owner, numFiles, totalSize, err :=
-		datasetIngestor.GetValidatedLocalFileList(datasetSourceFolder, datasetFileListTxt, symlinkCallback, filenameCheckCallback)
+	fullFileArray, err = prepareDataset(client, APIServer, user, originalMap, metaDataMap, tapecopies,
+		datasetSourceFolder, datasetFileListTxt, symlinkCallback, filenameCheckCallback)
 	if err != nil {
 		var emptyDatasetErr *datasetIngestor.EmptyDatasetError
 		var tooManyFilesErr *datasetIngestor.TooManyFilesError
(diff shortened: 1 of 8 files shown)


orchestrator/datasetIngestor.go
```diff
@@ -0,0 +1,105 @@
...
+	log.Println("File list collected.")
+	log.Printf("The dataset contains %v files with a total size of %v bytes.\n", numFiles, totalSize)
+
+	UpdateAndLogMetaData(client, APIServer, user, originalMap, metaDataMap, startTime, endTime, owner, tapecopies)
+	return fullFileArray, nil
+}
+
+// UpdateAndLogMetaData updates the dataset's metadata fields from the
+// scanned file list and logs the resulting metadata object.
+func UpdateAndLogMetaData(client *http.Client, APIServer string, user map[string]string,
```
omkar-ethz: (minor) maybe have it as a private function, as it doesn't seem to be useful outside this package.

I'd also delete the corresponding test, as it just repeats the test for the internal `datasetIngestor.UpdateMetaData` rather than mocking it, which is already tested in it's own package.

orchestrator/datasetIngestor.go
```diff
@@ -0,0 +1,142 @@
...
+// *datasetIngestor.EmptyDatasetError or *datasetIngestor.TooManyFilesError just mean this dataset
+// must be skipped (not fatal, no os.Exit); anything else is a hard failure gathering the local
+// file list. emptyDatasets/tooLargeDatasets are incremented to match whichever of those two
+// errors is returned.
+func PrepareDataset(client *http.Client, APIServer string, user map[string]string,
+	originalMap map[string]string, metaDataMap map[string]interface{}, tapecopies int,
+	datasetSourceFolder string, datasetFileListTxt string,
+	symlinkCallback func(symlinkPath string, sourceFolder string) (bool, error),
+	filenameCheckCallback func(filepath string) bool,
+	emptyDatasets *int, tooLargeDatasets *int) (fullFileArray []datasetIngestor.Datafile, err error) {
```
omkar-ethz, line 33: i'd remove these parameters `emptyDatasets *int` and `tooLargeDatasets *int`, as it mixes the caller's responsibility.

PrepareDataset already comminucates the EmptyDatasetError / TooManyFilesError, and the caller can update their state based on this accordingly.

orchestrator/datasetIngestor.go
```diff
@@ -0,0 +1,142 @@
...
+// ResolveCentralAvailability checks whether the dataset's source folder is available on the
+// central archive server via SSH, and decides the resulting copyFlag (currentCopyFlag is returned
+// unchanged when the data is centrally available).
+//
+// When the data is not centrally available, copying is required: on success (noninteractive, or
+// the user accepted via confirmContinue) it returns copyFlag=true alongside a
+// *NotCentrallyAvailableWarning - not a failure, just something the caller should report. It
+// returns ErrCopyRequiresPersonalAccount if no personal account (access group) is available, and
+// ErrIngestAborted if the user declines to continue.
+func ResolveCentralAvailability(username string, rsyncServer string, datasetSourceFolder string, sshOutput io.Writer,
```
omkar-ethz: (nitpick) maybe `sshOutput io.Writer` is not needed as a parameter, as it's not mocked in the tests (always nil). so `os.Stdout` can be passed directly to `datasetIngestor.CheckDataCentrallyAvailableSsh`

orchestrator/datasetIngestor.go
```diff
@@ -0,0 +1,142 @@
...
+		return currentCopyFlag, fmt.Errorf("cannot check if data is centrally available: %w", otherErr)
+	}
+	if sshErr == nil {
+		log.Println("Data is present centrally.")
+		return currentCopyFlag, nil
+	}
+
+	if len(accessGroups) == 0 {
+		return false, ErrCopyRequiresPersonalAccount
+	}
```
omkar-ethz: this can happen at the beginning of the function, so that the SSH check might be saved.

orchestrator/datasetIngestor_test.go
```diff
@@ -0,0 +1,350 @@
...
+		w.Write([]byte(`[]`))
+	}))
+	defer ts.Close()
+
+	folder := newTestDatasetFolder(t, map[string]string{"a": "hello"})
+
+	// GetLocalFileList walks each line of the file listing independently, so listing the same
+	// real file more times than TOTAL_MAXFILES allows drives numFiles past the limit without
+	// creating hundreds of thousands of files on disk (which is prohibitively slow, especially
+	// on Windows with real-time antivirus scanning).
```
omkar-ethz: General comment about the tests in this file:
Most of these (except `TestResolveCentralAvailability_*`) look like e2e tests, is that intended? if not, maybe this should be moved to `getLocalFileList_test.go` 

For unit tests, the datasetIngestor dependency should be mocked

minottic: Not intended as e2e, no real backend or remote host, just httptest faking the HTTP layer and a real temp dir for the filesystem side. I'd call these integration tests rather than unit tests: they deliberately exercise PrepareDataset together with the real GetLocalFileList, since what's being asserted (counters, error types, metadata mutation) depends on that real interaction.
That's different from ResolveCentralAvailability's tests, which mock datasetIngestor's SSH seams directly — but that's mostly pragmatic rather than principled: httptest.NewServer makes it cheap to spin up a real HTTP server to test against, and there's no equivalent shortcut for SSH, so mocking the SSH calls directly was the practical choice there.

TBH the intent here is simply to verify that combination of flags and counters as well as abort on errors work fine when the smaller pieces from datasetUtils and datasetIngestor are glued together to cover the business logic. I am happy to change it if you have suggestions on how to have test cases covered

orchestrator/datasetIngestor.go
minottic: done

orchestrator/datasetIngestor.go
minottic: done

orchestrator/datasetIngestor.go
minottic: done

orchestrator/datasetIngestor.go
minottic: rather than moving the increment logic into the cobra command, I split this into a private function (the actual dataset-preparation logic, no side effects on caller state) and a thin PrepareDatasetAndUpdateCounts wrapper that calls it and updates the counters. So the core logic stays free of the pointer out-params; only the wrapper touches shared state, and it exists in the orchestrator rather than the CLI command so it can be tested, refactored, and eventually reused elsewhere without pulling cobra into the dependency graph.

orchestrator/datasetIngestor_test.go
omkar-ethz: > to verify that combination of flags and counters as well as abort on errors work fine when the smaller pieces from datasetUtils and datasetIngestor are glued together

I was thinking unit tests by mocking the datasetIngestor functions: [here](https://github.com/paulscherrerinstitute/scicat-cli/commit/b7fac56c1883188194206bfce7ee53af2f2bbbef#diff-d3ba4d7998e47622b69c519e86c0c11759b58ec1f8e65708dd0bba2e4fdaedf2R15)

Then the [unit tests](https://github.com/paulscherrerinstitute/scicat-cli/commit/b7fac56c1883188194206bfce7ee53af2f2bbbef#diff-b3daa627674fc94794b81243f9f3888c3d9b14a49c2943db0f30efafda519c9bR16) for PrepareDatasetAndUpdateCounts still assert different outputs / errors, without needing the mock http backend, or to create so many files on disk. This runs in 4ms compared to 6s for the integration tests.

But maybe it's better to have the integ tests in addition to / instead of the unit tests, so your call!

omkar-ethz (approved):

orchestrator/datasetIngestor_test.go
minottic: ah I see, I think you are completely right, especially because this functions don't don't much that is not already tested in the mocked functions. I will take your commit, thanks!

PR #268, opened 2026-07-15
fix: append port when missing to rsync server

minottic:


datasetIngestor/scp.go
```diff
@@ -235,8 +235,18 @@ func (c *Client) walkAndSend(w io.Writer, src string) error {
 	return nil
 }
 
+// defaultSshPort appends the standard ssh port (22) to server if it doesn't already specify one.
+func defaultSshPort(server string) string {
+	if _, _, err := net.SplitHostPort(server); err != nil {
+		return server + ":22"
```
fpotier:
```suggestion
		return net.JoinHostPort(server, "22")
```
Probably not very common but this is more robust for ipv6

fpotier (approved):

PR #273, opened 2026-07-28
fix: set remote ingest as archivable

minottic:


orchestrator/datasetIngestor.go
omkar-ethz: could you update this explanation?

omkar-ethz (approved):

orchestrator/datasetIngestor.go
minottic: right, I did not change the comment, my bad, will do. As a general note, I moved to archivable=true as this will enable users to trigger from UI for data which is remote and Michi will simply check if origs are there and try to create otherwise

omkar-ethz: Thanks for the explanation!

OK - how does Michi create origs - with the completeIngest command?

minottic: yes. Should I add that to the comment? Sorry, I resolved and merged but should have waited for you to resolve it. I can do in another PR

omkar-ethz: No worries, it's just for my understanding, thanks!
You could add in the next PR:)

PR #275, opened 2026-07-29
feat: implement s3 transfer type for datasetIngestor

omkar-ethz:
Implement `--transfer-type s3` option for `datasetIngestor`.

### Example command:
```bash
go run ./cmd/ datasetIngestor --devenv --token $SCICAT_TOKEN --copy --transfer-type s3 --ingest ~/Downloads/metadata.json [filelisting.txt | folderlisting.txt]
```

After ingesting the dataset, the copy step uploads the sourceFolder to an s3 bucket mounted on Arema: psi-upload{-dev,-qa}. The credentials for upload are obtained from s3-broker by exchanging `SCICAT_TOKEN` for short term credentials

### Implementation notes
- Used [transfermanager](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager) which uses multipart uploads and goroutines to efficiently upload directories to S3
- The transfermanager client is initialized with short-term creds from s3-broker, refreshed automatically
- symlinks are skipped, as s3 does not have a related notion
- filelisting.txt support is implemented using transfermanager's [FileFilter](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager#FileFilter) callback. Windows file path quirks are taken into account.

### Manual testing

Tested following cases:
- nested directories
- symlinks
- filelisting.txt
- folderlisting.txt

Tested in QA that ingested and copied datasets can be archived by Michael's updated workflow.

<details>

<summary>
<h3> Design / unit tests </h3>
</summary>

At a high level, S3 transfer functionality has two dependencies - datasetIngestor's MarkFilesReady, and the core "upload directory to s3" functionality, abstracted in the `upload` function (which encompasses 1. credentials fetching for transfermanager and constructing a client, and 2. calling UploadDirectory on the client)
These two dependencies are collected in `s3Transfer` struct, and `transferFiles` is made a method on this struct. The tests for transferFiles then substitute mocks for these dependencies and verify control flow.

We also test the only other logic heavy method: `transferDirectory` by substituting a mock transfermanager client in tests, and asserting that the client's UploadDirectory is called with expected inputs (e.g. source dir / key prefix / file filter etc)

Similar to SshTransfer and GlobusTransfer, we export a single function TranferFilesS3 (which only sets up `s3transfer` with the real `upload` / `MarkFilesReady` functions and calls its transferFiles), that is assigned to transferFiles in cmd/datasetIngestor

</details>


cmd/cliutils/s3Transfer.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	output, err := client.UploadDirectory(ctx, input)
+	if err == nil {
+		log.Printf("Uploaded %v objects, failed %v objects\n", output.ObjectsUploaded, output.ObjectsFailed)
+	}
+	return err
+}
+
+// create a set from fileList, transforming to (linux style) absolute paths
+func makeAbsFilePathSet(sourceFolder string, fileList []string) map[string]struct{} {
```
minottic, line 147: minor: do these funcs belong here? They seem general file utils (same for FilterFile)

cmd/cliutils/s3Transfer.go
```diff
@@ -0,0 +1,187 @@
...
+
+type s3Creds struct {
+	AccessKey    string `json:"access_key"`
+	SecretKey    string `json:"secret_access_key"`
+	SessionToken string `json:"session_token"`
+	ExpiryTime   string `json:"expiry_time"`
+}
+
+// getShortTermCreds makes a GET request to brokerServer's /datasets/s3-creds endpoint with relevant params
+func getShortTermCreds(client *http.Client, brokerServer string, datasetId string, operation string, accessToken string) (s3Creds, error) {
```
minottic, line 164: minor: this does not seem to do too much but I wander if it is useful to have small test here

cmd/cliutils/s3Transfer.go
```diff
@@ -0,0 +1,187 @@
...
+	operation    string
+	accessToken  string
+}
+
+func (s *s3BrokerCredsProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
+	s3Creds, err := getShortTermCreds(s.client, s.brokerServer, s.datasetId, s.operation, s.accessToken)
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+	expires, err := time.Parse(time.RFC3339, s3Creds.ExpiryTime)
```
minottic, line 75: minor: worth returning if err, nil, err?

cmd/commands/datasetIngestor.go
```diff
@@ -100,6 +102,7 @@ For Windows you need instead to specify -user username:password on the command l
 		// globus specific vars (if needed)
 		var globusClient globus.GlobusClient
 		var gConfig cliutils.GlobusConfig
+		skipSymlinks := ""
 
 		switch transferType {
```
minottic, line 107: would the switch benefit from being moved to the orchestrator so it can be tested?

cmd/commands/datasetIngestor.go
```diff
@@ -421,6 +431,10 @@ For Windows you need instead to specify -user username:password on the command l
 							Filelist:       filePathList,
 							IsSymlinkList:  isSymlinkList,
 						},
+						S3Params: cliutils.S3Params{
```
minottic, line 434: I find it a bit weird that the S3Prams (as well as some fields of the others) get built inside the loop, even if they only need static params. What if the switch on top returned the class with all params filled in but the dynamic ones that are appended here? (and maybe moved to orchestrator for testing). It would also work to do in another PR

minottic (commented):
thanks! some minor comments, the only ones that I see more relevant are the missing tests ones

cmd/cliutils/s3Transfer.go
omkar-ethz: Now added a test for
1. getShortTermCreds, by mocking the http server
2. Retrieve, by mocking getShortTermCreds

cmd/cliutils/s3Transfer.go
omkar-ethz: good idea, returning empty, err in case of err explicitly

cmd/commands/datasetIngestor.go
omkar-ethz: i think this would be nice too, but seems like a big refactor, so will do it in another PR

cmd/commands/datasetIngestor.go
omkar-ethz: will attempt it in the next PR that includes moving the static params up to the switch

cmd/cliutils/s3Transfer.go
omkar-ethz: FilterFile is here to satisfy the transfermanager's FileFilter interface, and makeAbsFilePathSet is just a helper for it - i think they are quite transfermanager specific so leaving them here

minottic (approved):
thanks!

PR #276, opened 2026-08-05
feat: add landingzones from transfer type to job

minottic:
Extend transer-types to datasetArchiver, including the type rocrate and add them as landingZones to jobParams

### DIFF
--- a/cmd/commands/datasetArchiver.go
+++ b/cmd/commands/datasetArchiver.go
@@ -23,7 +23,10 @@ var datasetArchiverCmd = &cobra.Command{
 You must choose either an ownerGroup, in which case all archivable datasets
 of this ownerGroup not yet archived will be archived.
 Or you choose a (list of) datasetIds, in which case all archivable datasets
-of this list not yet archived will be archived. 
+of this list not yet archived will be archived.
+
+The transfer type used for the archival job's landing zone can be selected
+via the --transfer-type flag. Available options: "ssh", "globus", "rocrate".
 
 For further help see "` + cliutils.MANUAL + `"`,
 	Run: func(cmd *cobra.Command, args []string) {
@@ -48,6 +51,7 @@ For further help see "` + cliutils.MANUAL + `"`,
 		nonInteractiveFlag, _ := cmd.Flags().GetBool("noninteractive")
 		ownerGroup, _ := cmd.Flags().GetString("ownergroup")
 		showVersion, _ := cmd.Flags().GetBool("version")
+		transferType, _ := cmd.Flags().GetString("transfer-type")
 
 		if datasetUtils.TestFlags != nil {
 			datasetUtils.TestFlags(map[string]interface{}{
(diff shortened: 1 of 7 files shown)


omkar-ethz:
lgtm! could you try to merge main, and i will review the diff again?

cmd/commands/datasetArchiver.go
```diff
@@ -23,7 +23,10 @@ var datasetArchiverCmd = &cobra.Command{
 You must choose either an ownerGroup, in which case all archivable datasets
 of this ownerGroup not yet archived will be archived.
 Or you choose a (list of) datasetIds, in which case all archivable datasets
-of this list not yet archived will be archived. 
+of this list not yet archived will be archived.
+
+The transfer type used for the archival job's landing zone can be selected
+via the --transfer-type flag. Available options: "ssh", "globus", "rocrate".
```
omkar-ethz: also "s3"

omkar-ethz (approved):

cmd/commands/datasetArchiver.go
```diff
@@ -145,6 +160,7 @@ func init() {
 	datasetArchiverCmd.Flags().Bool("devenv", false, "Use development environment instead or production")
 	datasetArchiverCmd.Flags().Bool("noninteractive", false, "Defines if no questions will be asked, just do it - make sure you know what you are doing")
 	datasetArchiverCmd.Flags().String("ownergroup", "", "Specifies to which owner group should the archival job belong. If no dataset id's are passed, all datasets belonging to this ownergroup that can also be marked as archivable will be included. If not specified, a list of dataset id's must be passed as arguments instead")
+	datasetArchiverCmd.Flags().String("transfer-type", "ssh", "Selects the transfer type to be used for transferring files. Available options: \"ssh\", \"globus\", \"rocrate\"")
```
omkar-ethz: `+ "s3"`

cmd/commands/datasetIngestor.go
```diff
@@ -128,14 +128,16 @@ For Windows you need instead to specify -user username:password on the command l
...
+		case datasetUtils.S3:
 			transferFiles = cliutils.TransferFilesS3
 
 			if cmd.Flags().Changed("linkfiles") && linkfiles != "delete" {
 				log.Fatalln("Only --linkfiles=delete supported with transfer-type s3")
 			}
 			log.Println("WARNING: transfer-type is s3: symbolic links will be dropped from ingestion and copying")
 			skipSymlinks = "sA"
+		default:
+			log.Fatalf("unsupported transfer type for datasetIngestor: %q. Available options: \"ssh\", \"globus\", \"s3\"\n", transferTypeFlag)
```
omkar-ethz, line 140: `+ "rocrate"`

minottic: I actually left it out on purpose here. Unlike the other cases, rocrate does not do a copy for real, so adding the option on ingest feels a bit weird, as it would only work with data somehow already on the rocrate server, and it can be there already only if there was a http upload, for which only an archive would be needed and the rocrate service uses the datasetArchive separatly since it submits a single job for multiple datasets

omkar-ethz: Ok understood. So if i understand right - in the datasetArchiver command, the transfer-type option is a signal on which transfer-type _was_ used for this dataset, so that the correct landing zone can be set in the job.
Thanks for the explanation!

